PR #1, opened 2024-03-28
Linear Sum Agreement Algorithm

daniel-klein:
Experimenting with pair formation via linear sum agreement (LSA), as in EmbeddingNet.

Resulting pars are a bit different, LSA is on the right.
image

### DIFF
--- a/networks.py
+++ b/networks.py
@@ -12,6 +12,8 @@ Overview:
 import starsim as ss
 import numpy as np
 import pandas as pd
+import scipy.optimize as spo
+import scipy.spatial as spsp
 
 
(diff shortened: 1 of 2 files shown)


daniel-klein:
Note, LSA will create at most one partner per person per step, unlike the previous algorithm, so would be good to check concurrency.

daniel-klein:
image

daniel-klein:
Main takes 10.37 s and this branch is 19.59 s, so about 2x slower with LSA.

daniel-klein:
In the lifetime partners plot above, I'm surprise to see LSA making so many very high results. Not sure what's going on there!

robynstuart:
Closing this for now, can re-open once repo is up to date with latest starsim

PR #15, opened 2024-05-16
Hiv merge

alina-muellenmeister:
Brings the hiv module up to date with version 0.5.0 of Starsim


stisim/celery.py
```diff
@@ -0,0 +1,82 @@
+# FUNCTIONS TO RUN MODELS ON CLUSTER
```
robynstuart, line 1: I wonder whether this file should live here or somewhere else, e.g. the analyses folder... @alina-muellenmeister @RomeshA how have you done this with other work?

stisim/diseases/hiv_main.py
```diff
@@ -0,0 +1,204 @@
+"""
+Run hiv
```
robynstuart, line 2: What's the difference between this file and hiv.py?

stisim/multisim.py
```diff
@@ -0,0 +1,1235 @@
+import numpy as np
```
robynstuart, line 1: Maybe we can remove this an use ss.MultiSim?

robynstuart (commented):
Looks good! Just a couple of comments

stisim/multisim.py
alina-muellenmeister: oh yes, that sounds good!

stisim/diseases/hiv_main.py
alina-muellenmeister: `hiv.py` is the disease module, `hiv.main.py` is what I added for running simulations with different parameters (for calibration).` hiv.main.py `basically contains the simulation setup like in `run_hiv.y` and is called by the celery-module to set up the simulation with specific input params. Does that make sense? I am not sure if it should live in the disease-folder. For the gavi work, we had these types of files living in what would be the syphilis-analyses->stisim folder

stisim/celery.py
alina-muellenmeister: We had it together with the interventions and networks modules for the gavi work, but I am not sure either if it should live somewhere else

.gitignore
```diff
@@ -122,3 +122,92 @@ flake8_junit.xml
 
 .ipython
 .matplotlib
+figures/individual_viral_dynamics/viral_dynamics_agent_0.png
```
cliffckerr, line 125: I'd suggest just adding `*.png` rather than naming individual files

analyses/data/world_bank_hiv_prevalence.csv
```diff
@@ -0,0 +1,271 @@
+﻿"Data Source","World Development Indicators",
```
cliffckerr, line 1: This file isn't too large so it's not necessarily a problem to add it to the repo, but we should think about what data we want to add to the repo directly, vs what we want to store externally, or provide the script to download the data rather than the data itself (which in this case might be as simple as e.g. `sc.download('https://api.worldbank.org/v2/en/indicator/SH.DYN.AIDS.ZS?downloadformat=csv')`).

stisim/plotting.py
```diff
@@ -0,0 +1,84 @@
+####### PLOT RESULTS
+import matplotlib.pyplot as plt
+import numpy as np
+import stisim as ss
+from stisim import interventions as ssi
+import sciris as sc
+
+
+def plot_cum_infections(sim, ax, disease):
```
cliffckerr, line 9: If desired, these functions could be refactored to also take an argument for the result being plotted, so they could be combined into a single `plot_result()` function. Would be nice if `Result` objects had a `label` attribute though -- added an issue here: https://github.com/starsimhub/starsim/issues/510

cliffckerr (commented):
Looks good!

stisim/celery.py
RomeshA: It would normally be in the analyses folder - there is some commonality across analyses but separating that out seems like more trouble than it would be worth

stisim/diseases/hiv_main.py
RomeshA: I think this one would probably also be analysis-specific. The role of this file would be to create and configure a `Sim` for the analysis so I suspect this will eventually morph into something that sets up both the HIV and syphilis modules and associated interventions (so maybe it'll get renamed too)

RomeshA:
Looks good to me too, @alina-muellenmeister can we also add an automated test for the HIV module like we have for syphilis? (should just be a small number of agents and short timeframe, mainly to check that things load and run, rather than verifying results)

RomeshA:
Oh also, do we still need `run_starsim_hiv.py`? What does it do differently to `run_hiv.py`?

alina-muellenmeister:
Oh yes, `run_starsim_hiv.py` can be removed - I initially added it to understand what the hiv-module in starsim does, but can be deleted now!

.gitignore
RomeshA: Done - also since the scope of `.gitignore` files is everything under the directory, we can have multiple `.gitignore` files e.g., if we want to name specific files inside the `analyses` folder that are too specific for the main STIsim file - I've added an example of that too (cc @alina-muellenmeister)

alina-muellenmeister: Thank you, Romesh!

stisim/plotting.py
RomeshA: Made an issue for this

PR #20, opened 2024-05-22
Connector updates

RomeshA:
Updates to the connector structure - key differences are having `rel_sus` and `rel_trans` computed from scratch each timestep, so the connector can modify them freely. Also adds `SyphilisPlaceholder` to use when testing HIV without the actual syphilis module

Note that this requires the bug-fix in https://github.com/starsimhub/starsim/pull/520 (on the `pass-constructor-arguments` branch of Starsim)


RomeshA:
nb. key functionality is demonstrated in `devtest_hiv_syph.py` where syphilis coinfection increase HIV transmission

image

and see also `test_hiv.py` which contains an intervention and analyzer to explicitly control infection with HIV and syphilis, and start/stop times for ART, to confirm the time series for CD4 count, `rel_sus` and `rel_trans`. There are some discontinuities/edge cases at the moment that we might want to look more closely at later on

image

image

PR #22, opened 2024-06-03
Calib work

robynstuart:
- [ ] adds automatic calibration using optuna
- [ ] fixes a bug with testing (dt-adjustment was being applied twice)
- [ ] fixes a bug with hiv death probs (dt adjustment wasn't being applied)

Still WIP until calibration runs


analyses/run_hiv.py
```diff
@@ -2,93 +2,31 @@
 Run hiv
 """
 
+# Additions to handle numpy multithreading
```
robynstuart, line 5: these things shouldn't be there but have been removed in the repo-split branch

PR #27, opened 2024-06-16
Refactor lognormal

RomeshA:
Refactor some distributions to improve performance - the biggest change is that the previous implementation where the parameters for the dists directly called `get_age_risk_pars`, separate calls had to be made to compute `loc` and `scale` whereas in this implementation only one call is required


alina-muellenmeister (commented):
Looks good to me!

PR #42, opened 2024-07-03
Track syphilis transmission sources

robynstuart:
Track syphilis transmitters


robynstuart:
Might replace this later with an infection-log based approach, but for the moment this works, and new transmissions match new acquisitions

PR #43, opened 2024-07-03
Add extra metadata to infection log

RomeshA:
I was wondering about something like this, with the metadata captured in the infection log rather than adding new states. Then an analyser could use the log (e.g., via `df = log.line_list`) to aggregate up the various quantities, grouping by time etc., 

Opening this now for you to have a look at - one thing that is interesting is that the count of new infections (the number of rows in the log) is coming out as being slightly different to the sum of new infections recorded in the sim results. I'm not sure yet where that difference is coming from


robynstuart:
Closing for now since we don't need it, but will leave the branch so we can revisit if/when we do

PR #119, opened 2025-05-12
v1.1

robynstuart:
Version 1.1 of STIsim includes:
- improvements to the `Calibration` class: this class now inherits directly from the Starsim calibration class, so users will have access to easier parameter constraints, plotting, flexible fit functions, etc
- generalization of the coinfection class to handle any two diseases
- addition of a more detailed `BV` model courtesy of @lrd3uu and @jamiecohen's work in the VMB space

### DIFF
--- a/stisim/analyzers.py
+++ b/stisim/analyzers.py
@@ -23,24 +23,53 @@ class result_grouper(ss.Analyzer):
         return out
 
 
-# TODO: generalize this for any 2 diseases
 class coinfection_stats(result_grouper):
-    def __init__(self, diseases=None, *args, **kwargs):
+    """
+    Generates stats for the coinfection of two diseases.
+    This is useful for looking at the coinfection of HIV and syphilis, for example.
+
+    Args:
+        disease1 (str | ss.Disease): name of the first disease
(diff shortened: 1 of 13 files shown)


stisim/analyzers.py
```diff
@@ -23,53 +23,85 @@ def cond_prob(numerator, denominator):
...
+        # if the diseases are objects, get their names and store them instead of the objects
+        if isinstance(self.disease1, ss.Disease):
+            self.disease1 = self.disease1.name
+        if isinstance(self.disease2, ss.Disease):
+            self.disease2 = self.disease2.name
+
+        self.disease1_infected_state_name = disease1_infected_state_name
+        self.disease2_infected_state_name = disease2_infected_state_name
+        self.age_limits = age_limits or [15, 50]
+        default_denom = lambda self: (self.sim.people.age >= self.age_limits[0]) & (self.sim.people.age < self.age_limits[0])
```
jamiecohen, line 60: would it make sense to add in any sex limits -- ie presumably for bv-hiv youd only be looking at women

jamiecohen: ignore me i see you have it stratified by male and female below!

Issue #106, opened 2025-05-05
Redo folder structure for HIV


robynstuart:
Per [discussion on Slack](https://starsim-idm.slack.com/archives/C08B1MH4AA2/p1746134247669609), we are going to move HIVsim elements to a separate subfolder, as follows:

```
# Option 2
stisim/
stisim/pyproject.toml
stisim/stisim/__init__.py
stisim/stisim/diseases/
stisim/stisim/interventions/
stisim/stisim/networks.py
stisim/stisim/[...]
stisim/hivsim/__init__.py
stisim/hivsim/interventions.py
stisim/hivsim/networks.py
stisim/hivsim/[...]
```

robynstuart:
Decided against this for now; can reopen as needed

PR #115, opened 2025-05-08
Moved HIV into its own directory separate from the other STIs

RHull-IDM:
Closes #106

PR #121, opened 2025-05-12
Moved HIV into its own directory separate from the other STIs

robynstuart:
Reopens https://github.com/starsimhub/stisim/pull/115 against rc11
Closes https://github.com/starsimhub/stisim/issues/106


robynstuart:
@RHull-IDM thanks, this looks good! Hoping to get @cliffckerr's thoughts too. We're balancing a trade-off between having the HIV components more externally visible, vs adding complexity from treating them differently to the other STIs. I don't know whether I'm 100% convinced this is the right path, but hopefully it'll get easier once we add an `hivsim.Sim` class.

robynstuart:
Thank you @RHull-IDM! Upon further reflection we've decided to hold off on this for the moment - we can revisit down the track if needed.

Issue #107, opened 2025-05-05
Benchmark


robynstuart:
Add a test similar to https://github.com/amath-idm/hpvsim/blob/main/tests/test_baselines.py#L91 and https://github.com/amath-idm/hpvsim/blob/main/tests/test_baselines.py#L108 to record sim run time and results

PR #123, opened 2025-05-16
Added test to track baselines outputs and performance benchmarks.

RHull-IDM:
Addresses #107.


RHull-IDM:
The baseline includes all of the diseases and some of the connectors and interventions, but not all of them. I tried to find a good balance between coverage and not totally overloading the test with compute load, but that means changes to the components not included won't necessarily break this test. Would you rather keep it like this, or make it exhaustive?

robynstuart:
@RHull-IDM this looks great to me! Would you mind re-opening this PR against the `rc12` branch instead?

RHull-IDM:
@robynstuart base branch changed

robynstuart:
I hadn't realized how long the baseline.json file is... I think it might be difficult to meaningfully compare sims. We might need to first determine the core metrics that we would want to compare, and then only save a subset of results. I'll leave this open for now and we can iterate on the approach.

RHull-IDM:
Todo: look for ways to summarize sim results to compare instead of every single sim output. Maybe like FPsim?

robynstuart:
Closing for now, will revisit when we have a more parsimonious summary

PR #146, opened 2025-07-01
Added sim init logic

RHull-IDM:
Adds an sti.Sim() class which handles defaults for most sim modules and provides greater flexibility in specifying inputs to the sim.


robynstuart:
Merging despite the failing tests, which are starsim v3 issues

PR #154, opened 2025-08-12
Starsim v3 checks

robynstuart:
Added examples of failing tests - looks like time conversion needs a little work to get these to pass - @cliffckerr welcome inputs from or your robot stand-in!


robynstuart:
Merging, I think we're getting closer!

PR #167, opened 2026-02-04
Add HIVsim skeleton

cliffckerr:
Proof of principle HIVsim implementation

### DIFF
--- a/hivsim/sim.py
+++ b/hivsim/sim.py
@@ -5,11 +5,17 @@ import stisim as sti
 from stisim.diseases.hiv import HIV, HIVPars
 from stisim.interventions.hiv_interventions import HIVTest, ART, VMMC, Prep
 
-__all__ = ['Sim', 'HIV', 'HIVPars', 'HIVTest', 'ART', 'VMMC', 'Prep']
+__all__ = ['Sim', 'HIV', 'HIVPars', 'HIVTest', 'ART', 'VMMC', 'Prep', 'ss', 'sti']
 
 class Sim(sti.Sim):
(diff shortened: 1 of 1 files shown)


hivsim/sim.py
```diff
@@ -0,0 +1,13 @@
+import starsim as ss
+import stisim as sti
+
+class Sim(sti.Sim):
```
robynstuart: This is great. Total aside but I wonder how hivsim should be imported. `import hivsim as hiv` might lead to chaos if we also typically call the disease module `hiv`.

cliffckerr: Yes I think `import hivsim` is probably fine ... or `import hivsim as hs`? Although that seems weird and unnecessary. What will people be using it for other than the sim?

tests/test_hiv.py
```diff
@@ -160,14 +161,15 @@ def test_hiv():
...
+    pars['init_prev'] = 0
+    pars['p_hiv_death'] = 0
+    pars['include_aids_deaths'] = False
     pars['demographics'] = [ss.Pregnancy(fertility_rate=0), ss.Deaths(death_rate=0)]
     pars['interventions'] = PerformTest(events)
     output = TrackValues()
     pars['analyzers'] = output
 
-    sim = sti.Sim(pars).run()
+    sim = hivsim.Sim(pars).run()
```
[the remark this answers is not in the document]
cliffckerr: Fixed for now since it was failing due to a bug in argument passing, but I agree we should write better tests soon!

Issue #181, opened 2026-02-12
Create repo structure


robynstuart:
stisim/
├── hivsim/
│   ├── sim.py
│   ├── locations.py
│   └── data/
│       └── generic/        # Minimal defaults only
│           └── parameters.py

# Official maintained locations (optional - 1-3 key examples)

stisim/
├── hivsim/
│   └── data/
│       ├── generic/
│       └── example_zambia/  # Single well-documented example

# External locations (separate repos)
hiv_kenya/
hiv_zambia/
hiv_zim/
hiv_kenya_who/
hiv_southafrica/
... (etc)

PR #185, opened 2026-02-16
HIV reorg

robynstuart:
Reorganize repo structure to create minimal examples. Still WIP; does this structure make sense?

See changelog and readme for use patterns and change summary

Closes https://github.com/starsimhub/stisim/issues/181


stisim_examples/loaders.py
```diff
@@ -0,0 +1,122 @@
+"""
+Data loading utilities for location-specific STIsim examples.
+"""
+import pandas as pd
+import sciris as sc
+from pathlib import Path
```
cliffckerr: Can use `sc.path` to avoid this import if you prefer

stisim/plotting.py
```diff
@@ -0,0 +1,258 @@
...
+# After validate_sim_data, the time column becomes the index,
+# and the data columns may use dot or underscore notation.
+_DATA_COL_MAP = {
+    'hiv_new_infections': ['hiv.new_infections', 'hiv_new_infections'],
+    'hiv_new_deaths':     ['hiv.new_deaths', 'hiv_new_deaths'],
+    'hiv_n_infected':     ['hiv.n_infected', 'hiv_n_infected'],
+    'hiv_prevalence':     ['hiv.prevalence', 'hiv_prevalence'],
+    'hiv_n_on_art':       ['hiv.n_on_art', 'hiv_n_on_art'],
+    'n_alive':            ['n_alive'],
+}
```
cliffckerr: I think we could use a function for this instead using a regex that matches either `_` or `.`

cliffckerr (commented):
First reaction is that I think for something with as wide a scope as STIsim, it would probably make sense to have a separate data repo, assuming we want to pre-download and process the data files. Otherwise I'd be worried that this repo would become very large tracking different versions of everything from age distributions over time to syphilis prevalence.

The exception would be if we want to write data loaders that work on the fly, eg download and process all the data we need from publicly available sources. I don't love this because they become fragile (e.g. WorldPop changes their API URL, as they did before, and suddenly nothing works).

Examples:
```py
#%% STIsim simple API -- load automatically
sim = sti.Sim(location='kenya') # load all available kenya data

# STIsim Sim under the hood
if location is not None:
  data = sti.data.load(location)
  self.data = sti.Data(data)

# stisim.data implementation
def load(location):
  location = normalize_location(location) # Kenya -> kenya etc
  local_path = make_local_path(location)
  if not check_downloaded(local_path):
    url = 'https://github.com/starsimhub/stisim_data/raw/{location}.zip'
    sc.download(url, path=local_path) # download on use, not when stisim is installed
  data = sc.unzip(local_path)
  return data


#%% STIsim complex API -- load manually
sti.data.download_data(folder='./my_kenya_data')
# ... manually check/modify the data files ...
sim = sti.Sim(data='./my_kenya_data') # load data from folder

# STIsim Sim under the hood
if isinstance(data, str):
  data = sti.data.load_local(data)
  self.data = sti.Data(data)
```

stisim_examples/__init__.py
```diff
@@ -0,0 +1,158 @@
...
+        >>> sim = stx.HIVSim(location='kenya', dur=50)
+        >>> sim.run()
+    """
+    if location is None:
+        raise ValueError("location parameter is required. Use one of: " + ", ".join(LOCATIONS.keys()))
+
+    return Sim(demographics=location, diseases='hiv', **kwargs)
+
+
+def MultiSim(demographics=None, diseases=None, n_runs=10, **kwargs):
```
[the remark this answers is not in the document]
robynstuart: ah yes can remove, was just there while debugging

stisim_examples/demo/README.md
```diff
@@ -0,0 +1,55 @@
...
+import stisim_examples as stx
+
+# Create demo HIV simulation
+sim = stx.Sim(demographics='demo', diseases='hiv')
+sim.run()
+
+# Or use HIV-specific convenience function
+sim = stx.HIVSim(location='demo')
+sim.run()
+```
```
ckirkman-IDM: Maybe it is just me, but it seems weird to have an example with no example/project-specific execution script or a call to a built-in script (for example, "python -m stisim.run <args_here>" )

stisim_examples/loaders.py
```diff
@@ -0,0 +1,122 @@
...
+        raise ValueError(f"Location '{location}' not found. Available locations: {available}")
+
+    # Normalize diseases to list
+    if isinstance(diseases, str):
+        diseases = [diseases]
+    elif diseases is None:
+        diseases = []
+
+    # Get location path
+    loc_path = Path(__file__).parent / location
```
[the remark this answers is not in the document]
ckirkman-IDM: One detail that may be worth considering is that ".parent" can return the current path ('.') if the path you are getting the parent for is not a full path (for example, ./get_parent_for_this_file.py), which can cause issues. This can definitely be wrong. Might want to expand out your path to a full path before parent, just in case (not sure what python defaults for __file__ are, but who wants to remember? :P)

stisim_examples/README.md
```diff
@@ -0,0 +1,247 @@
...
+sim.run()
+```
+
+### Pattern 3: Via hivsim_examples
+
+```python
+import hivsim_examples as hx
+
+# Same as stx.HIVSim
+sim = hx.Sim(location='zimbabwe')
```
ckirkman-IDM: missing:
```
, n_agents=5000, dur=40
```

?

stisim_examples/README.md
```diff
@@ -0,0 +1,247 @@
+# STIsim Examples
```
ckirkman-IDM: I've read through this file and my impression is that there are too many suggested/supported ways to do the same/nearly the same thing. It is one thing to provide flexibility, but reading the few characters-difference between the examples below requires some eye-squinting and triple-reading for context. This does not strike me as "low surprise" patterning, unfortunately (is sti in my import?, hiv? examples? core code? other? But they all look pretty much the same so closely examining the import, the slight syntax changes induced by the imported variant of (hiv)(sti)Sim, ...

I admit that I am a fan of having a clear, spelled-out, single usage pattern for loading project-type data (to be as unambiguous as possible).

Other notes, the difference between pattern 1 and 4 isn't explicitly spelled out (I think it is intervention-based).

stisim_examples/plotting.py
```diff
@@ -0,0 +1,8 @@
+"""
+Plotting functions for stisim_examples.
+
+Re-exports from stisim.plotting for convenience.
+"""
+from stisim.plotting import plot_hiv
+
+__all__ = ['plot_hiv']
```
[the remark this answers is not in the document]
robynstuart: Agreed, have removed and replaced with a few enhancements to ss.Sim().plot()

PR #296, opened 2026-02-23
Test to ensure untreated hiv cd4 counts cannot go up. (Fix #221) Test to ensure proper median time to AIDS (Fix #222)

### LINKED ISSUE #221: Test CD4 counts decline over time without treatment
Sub-issue of #193.

### LINKED ISSUE #222: Test median time from infection to AIDS
Sub-issue of #193.

ckirkman-IDM:
Adding new natural history test file & analyzers.

Addresses:
https://github.com/starsimhub/stisim/issues/221
https://github.com/starsimhub/stisim/issues/222

Question: Should CD4 < (or <= 200) be added as a condition in defining AIDS (in addition to falling state, or instead of)?

Update: I have updated this PR in response to CR comments, mostly for style, over at:
https://github.com/starsimhub/stisim/pull/302
- unittest -> pytest (format and assert updates)
  - I mostly like the switch away from the larger hodge-podge assert functions in unittest (assert BOOL, instead of assertEqual, assertGreaterEqual, assertLess, assertTrue, ... Readability of these basic asserts is better.
- added module level doc string
- added sciris import/timer/headings
- added descriptive assert-fail messages in addition to the descriptive comments.
  - I do like this standard.
- returning sim from tests
- added "if __name__ == "__main__"" block for direct "python xyz.py" execution
  - \_\_main__ standard can make for incomplete local-run tests unless we never make a mistake in adding new tests to the main block, but the output does look better than "python -m pytest xyz.py"
- doc strings on classes (analyzers)

I will push back on a couple of points:
- long test function names: These read like sentences intentionally, so that one does not need to read any code to know what the test is about. One COULD say this is the domain of a doc string, but those tend to be longer and not reported everywhere (and as a dev, my eyes first go to function name). I would not use function names this verbose in model/core behavior, but in tests I feel the extra letters/context outweighs the minor cost of length. Casing used is as per the style-guide (lowercase/'_' style)
  - for example, one might write multiple cd4-related tests (we probably will) and "def test_cd4" cannot do all of them and test just one aspect, also per the style guide. Context is needed.
- Analyzer classes vs basic functions
  - I feel this suggestion misses the full scope of the issue. It is really Analyzer classes vs "global variables and functions". While the latter may be shorter for very small analysis, it is less portable, reusable, may lead to future problems of shared data, and is in general a pattern I would not want to encourage for user. I feel it would be better to create an equivalent "TrivialAnalyzer" class to subclass ss.Analyzer that is super succinct and still stores results "in context" (on the analyzer object, not a global (or module) variable). I have had to deal with numerous problems with prior code when this global/module variable approach was taken once the code was reused.
  - ss.Analyzer is a built-in starsim format/mechanism and these tests also demonstrate making/using analyzer objects (equivalent gain as demonstrating simple function/global param use)

### DIFF
--- a/hivsim_examples/zimbabwe/sim.py
+++ b/hivsim_examples/zimbabwe/sim.py
@@ -17,7 +17,7 @@ import stisim as sti
 datadir = sc.thispath()
 
 # Sim settings
-sim_pars = dict(start=1990, stop=2025)
+sim_pars = dict(start=1990, stop=2025, age_scale=1000)
 
 # Calibrated parameters
 sti_pars = dict(
--- a/stisim/parameters.py
+++ b/stisim/parameters.py
@@ -41,6 +41,7 @@ class SimPars(ss.SimPars):
         self.birth_rate = None
         self.death_rate = None
         self.use_aging  = False
+        self.age_scale = 1.0
         self.use_pregnancy = True
         self.use_migration = False
 
--- a/stisim/sim.py
+++ b/stisim/sim.py
@@ -290,7 +290,7 @@ class Sim(ss.Sim):
             if self.pars['total_pop'] is not None:
                 total_pop = int(self.pars['total_pop'])
             else:
-                total_pop = int(age_data.value.sum() * 1000)  # Age data values are in thousands
+                total_pop = int(age_data.value.sum() * self.pars['age_scale'])  # Age data values are in thousands
             age_data['value'] /= sum(age_data['value'])  # Normalize the age distribution
             people = ss.People(self.pars.n_agents, age_data=age_data)
 


tests/pytest.ini
```diff
@@ -1,5 +1,5 @@
 [pytest]
 filterwarnings =
     ignore::pytest.PytestReturnNotNoneWarning
-env =
-    SCIRIS_BACKEND=agg
\ No newline at end of file
+# env =
+#    SCIRIS_BACKEND=agg
```
robynstuart, line 5: think we need to uncomment this again

tests/test_hiv_natural_history_verification.py
```diff
@@ -0,0 +1,101 @@
...
+verbose = False
+do_plot = False
+sc.options(interactive=False)
+
+
+@sc.timer()
+def test_cd4_counts_decline_over_time_without_treatment():
+    sc.heading("Ensuring CD4 counts decline without treatment")
+
+    sim = build_testing_sim(analyzers=[CD4ByUIDTracker()], n_agents=5, duration=5)
```
robynstuart: could be sim = sti.Sim(diseases='hiv', n_agents=5, dur=5, analyzers=[CD4ByUIDTracker()])

tests/test_hiv_natural_history_verification.py
```diff
@@ -0,0 +1,101 @@
...
+from testlib import build_testing_sim
+
+
+verbose = False
+do_plot = False
+sc.options(interactive=False)
+
+
+@sc.timer()
+def test_cd4_counts_decline_over_time_without_treatment():
```
robynstuart: still very long names 😅 what about `test_untreated_cd4`?

tests/test_hiv_natural_history_verification.py
```diff
@@ -0,0 +1,101 @@
...
+        assert minimum_delta <= 0, f"CD4 erroneously went up for at least one untreated agent"
+        if not found_decrease and minimum_delta < 0:
+            found_decrease = True
+    # ensure we have at least ONE agent with an actual decreasing CD4 (not just stable)
+    assert found_decrease is True, f"CD4 is stable for all untreated agents, no decrease observed"
+    return sim
+
+
+@sc.timer()
+def test_median_time_from_infection_to_aids_without_treatment():
```
robynstuart: `test_untreated_time_to_aids` ?

Note, it might be useful to add a `ti_aids` property to the HIV module - will make an issue

tests/test_hiv_natural_history_verification.py
```diff
@@ -0,0 +1,101 @@
...
+    if verbose:
+        print(f"{len(times_to_aids)} agents progressed to AIDS.")
+        print(f"median time to AIDS stage: {median_time} timesteps.")
+        print(f"expected_median: {expected_median} timesteps.")
+        print(f"min time to AIDS stage: {min(times_to_aids)} timesteps.")
+        print(f"mean time to AIDS: {mean(times_to_aids)} timesteps.")
+        print(f"max time to AIDS stage: {max(times_to_aids)} timesteps.")
+    delta = result_tolerance * expected_median
+    msg = f"median time to AIDS: {median_time} not within delta: {delta} of expected: {expected_median}"
+    assert median_time == pytest.approx(expected=expected_median, rel=result_tolerance), msg
```
robynstuart: interesting that this fails! Do you know what's going on? If not I'll make an issue to investigate

tests/pytest.ini
ckirkman-IDM: I commented this because pytest warned on the command line with env active (unknown parameter). Let me know if you still want it uncommented:

```
..\..\env-stisim\lib\site-packages\_pytest\config\__init__.py:1428
  C:\Users\clarkki\code\starsim\env-stisim\lib\site-packages\_pytest\config\__init__.py:1428: PytestConfigWarning: Unknown config option: env

    self._warn_or_fail_if_strict(f"Unknown config option: {key}\n")

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

tests/test_hiv_natural_history_verification.py
ckirkman-IDM: Details, details... :)  One thing I am working on is reducing the test size/scope within reason. There is no explicit need for a network here, so yes, it can be omitted. I tend to leave more stuff in than is strictly necessary to help ensure there is a lower likelihood of strange-interaction-type bugs.

However, if we stay on top of test coverage on various components + having good integration tests, there is less need for tests like this to bridge the gap (that I am, sadly, used to always existing at some level). I will update this and future tests in "natural history" to minimize components used (test dependent).

tests/test_hiv_natural_history_verification.py
ckirkman-IDM: Just to be clear, is "AIDS" == 'falling'? Or also CD4 < 200? . For this test it isn't critical (time to falling regression is good). But for an actual property, the definition is important (ti_aids could be dependent on CD4 fall rates, if it varies by individual)... (and treatment on/off/on/off).

tests/test_hiv_natural_history_verification.py
ckirkman-IDM: As discussed, finicky median vs mean. Updating to mean shortly...

tests/test_hiv_natural_history_verification.py
ckirkman-IDM: Ok, you're right, my names are a BIT long. Shortening some :) .

PR #297, opened 2026-02-24
Add hivsim examples and demo()

robynstuart:
This is a much less ambitious version of getting a demo set up, which doesn't require the heavy data refactoring of my previous approach (#185). Let me know what you think.

## Summary

- Add `hivsim_examples` package with two pre-configured HIV examples (`simple` and `zimbabwe`)
- Add `hs.demo()` function for quickly running examples (e.g. `hs.demo('zimbabwe')`)
- Add `Sim.plot()` override that auto-selects curated HIV keys (new infections, deaths, PLHIV, prevalence, ART, population)
- Add `data` parameter to `sti.Sim` for UNAIDS data overlay on plots
- Fix population scaling in `process_demographics()` (age data values are in thousands)
- Fix `separate_pars()` so kwargs override `sim_pars` defaults
- Fix `get_age_distribution()` to handle CSVs without a `year` column

## Usage

```python
import hivsim as hs

# Quick demo
hs.demo('zimbabwe')

# Customize
sim = hs.demo('zimbabwe', run=False, n_agents=5000)
sim.run()
sim.plot(annualize=True)

# Direct make_sim access
from hivsim_examples.zimbabwe.sim import make_sim
sim = make_sim(n_agents=5000)
```

### DIFF
--- a/hivsim/sim.py
+++ b/hivsim/sim.py
@@ -2,10 +2,7 @@ import sciris as sc
 import starsim as ss
 import stisim as sti
 
-from stisim.diseases.hiv import HIV, HIVPars
-from stisim.interventions.hiv_interventions import HIVTest, ART, VMMC, Prep
-
-__all__ = ['Sim', 'HIV', 'HIVPars', 'HIVTest', 'ART', 'VMMC', 'Prep', 'ss', 'sti', 'demo']
+__all__ = ['Sim', 'demo']
 
 class Sim(sti.Sim):
     """
@@ -27,7 +24,7 @@ class Sim(sti.Sim):
         sim_pars = sc.mergedicts(sim_pars)
         hiv_pars = sc.mergedicts(hiv_pars)
         default_sim_keys = ss.SimPars().keys()
-        default_hiv_keys = HIVPars().keys()
+        default_hiv_keys = sti.HIVPars().keys()
 
         # Pull modules out for special processing
         modules = sc.objdict()
@@ -56,7 +53,7 @@ class Sim(sti.Sim):
 
         # Handle interventions
         if not modules.interventions:
-            modules.interventions = [HIVTest(), ART(), VMMC(), Prep()]
+            modules.interventions = [sti.HIVTest(), sti.ART(), sti.VMMC(), sti.Prep()]
 
         super().__init__(
             pars          = sim_pars,
@@ -96,23 +93,21 @@ def demo(example=None, run=True, plot=True, **kwargs):
         hs.demo('zimbabwe')                            # Run Zimbabwe HIV model
         sim = hs.demo('zimbabwe', run=False, n_agents=500)  # Just create it
     """
-    import importlib
-
     if example is None:
         example = 'simple'
     if example not in EXAMPLES:
         available = ', '.join(EXAMPLES.keys())
         raise ValueError(f"Example '{example}' not found. Available: {available}")
 
-    mod = importlib.import_module(f'hivsim_examples.{example}.sim')
+    mod = sc.importbyname(f'hivsim_examples.{example}.sim')
     sim = mod.make_sim(**kwargs)
 
     if run:
         sim.run()
         if plot:
             try:
-                sim.plot(annualize=True)
+                sim.plot('hiv', annualize=True)
             except Exception:
-                sim.plot()
+                sim.plot('hiv')
 
     return sim
--- a/stisim/sim.py
+++ b/stisim/sim.py
@@ -192,26 +192,27 @@ class Sim(ss.Sim):
 
     def plot(self, key=None, **kwargs):
         """
-        Plot sim results. If HIV is present and no key is specified, shows a
-        curated 2x3 HIV panel (new infections, deaths, PLHIV, prevalence,
-        ART, population). Otherwise falls back to the standard starsim plot.
+        Plot sim results. If key is the name of a disease in the sim (e.g. 'hiv'),
+        shows a curated panel for that disease. Otherwise falls back to the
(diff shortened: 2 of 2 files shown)


hivsim/sim.py
```diff
@@ -67,4 +66,53 @@ def __init__(self, pars=None, sim_pars=None, hiv_pars=None, location=None, **kwa
             interventions = modules.interventions,
             **pars
         )
-        return
\ No newline at end of file
+        return
+
+
+# Available examples for hs.demo()
```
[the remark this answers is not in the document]
RHull-IDM: I didn't know we had a demo function in Starsim core, but consistency is good.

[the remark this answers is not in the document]
cliffckerr: Demo is just a sim with extra defaults enabled, so I think it makes sense to keep it here. But I can see how other choices would also make sense!

hivsim/sim.py
```diff
@@ -5,7 +5,7 @@
 from stisim.diseases.hiv import HIV, HIVPars
 from stisim.interventions.hiv_interventions import HIVTest, ART, VMMC, Prep
 
-__all__ = ['Sim', 'HIV', 'HIVPars', 'HIVTest', 'ART', 'VMMC', 'Prep', 'ss', 'sti']
+__all__ = ['Sim', 'HIV', 'HIVPars', 'HIVTest', 'ART', 'VMMC', 'Prep', 'ss', 'sti', 'demo']
```
cliffckerr: Why are these things being imported into the namespace from here, rather than from the original files? Also as I commented on the other PR, you can just do
```py
import stisim as sti
```
and then `sti.HIVTest`, etc.

hivsim/sim.py
```diff
@@ -67,4 +66,53 @@ def __init__(self, pars=None, sim_pars=None, hiv_pars=None, location=None, **kwa
...
+    """
+    import importlib
+
+    if example is None:
+        example = 'simple'
+    if example not in EXAMPLES:
+        available = ', '.join(EXAMPLES.keys())
+        raise ValueError(f"Example '{example}' not found. Available: {available}")
+
+    mod = importlib.import_module(f'hivsim_examples.{example}.sim')
```
cliffckerr: Can use `sc.importbyname(f'hivsim_examples.{example}.sim')` and avoid the `importlib` import. But that too feels hacky ... should we write a function to do this? e.g.
```py
example = hivsim_examples.get_example(example)
sim = example.make_sim(**kwargs)
```

cliffckerr: And then `hivsim_examples.get_example(example)` can basically just be `getattr(hivsim_examples, example)` ... so .... maybe we can just do that here lol

hivsim_examples/simple/sim.py
```diff
@@ -0,0 +1,11 @@
+"""
+Minimal HIV simulation example using hivsim defaults.
+"""
+import hivsim as hs
```
cliffckerr, line 4: Update import convention

stisim/sim.py
```diff
@@ -189,6 +190,30 @@ def init(self, force=False, **kwargs):
...
+        if key is None and 'hiv' in self.diseases:
+            key = [
+                'hiv_new_infections',
+                'hiv_new_deaths',
+                'hiv_n_infected',
+                'hiv_prevalence',
+                'hiv_n_on_art',
+                'n_alive',
+            ]
+        return super().plot(key=key, **kwargs)
```
cliffckerr, line 216: Hmm ... I don't love this, there's no easy way to access the regular plot functionality if you have HIV in your sim! How about one of:
- Doing this iff `key='hiv'` (can raise an exception if no HIV module, but honestly I think anyone trying to plot HIV without an HIV module will quickly realize their mistake anyway, so probably overkill)
- Calling this method `sim.plot_hiv()`
- Adding this to the HIV module, so users can do `sim.diseases.hiv.plot()`

cliffckerr: Or ... combining the first and last? If the user calls `sim.plot(key=known_disease)`, automatically call `sim.diseases[known_disease].plot()`

tests/test_examples.py
```diff
@@ -0,0 +1,55 @@
+"""
+Tests for hivsim_examples package and hs.demo()
+"""
+import numpy as np
+import hivsim as hs
```
cliffckerr, line 5: Update import

pyproject.toml
```diff
@@ -57,5 +57,8 @@ dev = [
 where = ["."]
 include = ["stisim*", "hivsim*"]
 
+[tool.setuptools.package-data]
+"hivsim_examples.zimbabwe" = ["*.csv"]
```
cliffckerr, line 61: Can you just do a wildcard for all CSV files? (Not sure if that works, should check)

cliffckerr (commented):
Looks good, some suggestions

hivsim_examples/simple/sim.py
robynstuart: addressed https://github.com/starsimhub/stisim/pull/299

tests/test_examples.py
robynstuart: addressed https://github.com/starsimhub/stisim/pull/299

stisim/sim.py
robynstuart: fixed

hivsim/sim.py
robynstuart: done

hivsim/sim.py
robynstuart: done

PR #300, opened 2026-02-24
Fix disease double-stepping bug (rel_trans for acute HIV)

robynstuart:
## Summary

- **Bug**: `self.dt = 'month'` in `BaseSTIPars` caused starsim's loop planner to schedule disease steps twice per timestep when the sim's dt matched. The second `step_state` call reset `rel_trans` to 1.0, nullifying the acute transmission boost (~6x) and causing incorrect transmission dynamics.
- **Fix**: Remove the explicit `dt` from `BaseSTIPars` so diseases inherit the sim's dt, preventing double-stepping.
- **New**: `aids` property on HIV class (`infected & cd4 < 200`) for convenient access.
- **Test**: `test_transmission.py` regression test verifying rel_trans values for acute (~6x), latent (~1x), and AIDS (~8x) stages.

### DIFF
--- a/stisim/diseases/hiv.py
+++ b/stisim/diseases/hiv.py
@@ -252,7 +252,7 @@ class HIV(BaseSTI):
 
     @property
     def aids(self):
-        return self.infected & (self.cd4 < 200)
+        return self.cd4 < 200
 
     def make_p_hiv_death(self, uids=None):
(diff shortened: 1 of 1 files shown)


cliffckerr (changes requested):
Couple changes sugged

PR #302, opened 2026-02-25
Test latent rel_trans is always 1 and acute rel_trans is always > 1 (after infection timestep) (Fix #226)

### LINKED ISSUE #226: Test acute phase has elevated transmission
Sub-issue of #194.

ckirkman-IDM:
https://github.com/starsimhub/stisim/issues/226


robynstuart:
Hey @ckirkman-IDM, thanks for working on this!!! A few things to bring in line with our [style guide](https://github.com/starsimhub/styleguide) before merging:

1. Simplicity simplicity simplicity!
Cliff gave an example previously demonstrating how this test (an 86-line analyzer class + a test class with setUp + the test itself) is almost equivalent to a 10-line version:

```
import stisim as sti

acute_rel_trans = []
def acute_tracker(sim):
    hiv = sim.diseases.hiv
    prior_acute = hiv.acute & (hiv.ti_infected < sim.ti)
    acute_rel_trans.extend(hiv.rel_trans[prior_acute].tolist())

sim = sti.Sim(dur=10, n_agents=100, diseases=sti.HIV(), analyzers=acute_tracker)
sim.run()
assert min(acute_rel_trans) > 1, 'Expected all acute rel_trans > 1'
```

Starsim analyzers accept plain functions so there's no need for a class unless you're doing something genuinely complex. I realize your test here IS more complex, but See PR #300 (test_transmission.py) for a recent example that's slightly more involved but still keeps things minimal. 

2. Use plain functions, not unittest.TestCase
Our tests need to work both as pytest tests and as standalone scripts (style guide §3). The unittest.TestCase class pattern breaks this dual-mode convention. Tests should be plain test_*() functions with an if __name__ == '__main__': block for standalone execution. See test_sim.py for the canonical pattern. (ref: [3_tests.md on File Structure](https://github.com/starsimhub/styleguide/blob/main/3_tests.md))

3. No type annotations
Our style guide departs from Google on type annotations (GSG 2.21). We don't use them. Starsim inputs are intentionally flexible (a parameter might accept a scalar, list, or array), so annotations tend to be misleading. This applies to build_testing_sim(diseases: list, n_agents: int = 500, ...) in testlib.py and states: list[str] in the analyzer. Document types in docstrings instead.

4. Missing standard test boilerplate
Each test file needs:
- Module-level docstring ("""Test HIV natural history verification""")
- import sciris as sc and sc.options(interactive=False)
- @sc.timer() decorator on each test function
- sc.heading('Testing ...') at the top of each test
- Return the sim for interactive exploration
- if __name__ == '__main__': block

5. Use assert statements, not self.assert*()
Per the style guide, use plain assert with descriptive messages:
```
# Instead of:
self.assertGreater(len(acute_ratios), 0)
# Use:
assert len(acute_ratios) > 0, 'Expected at least one acute ratio'
```

6. Docstrings, not comments, on classes
The analyzers use # Records the time to AIDS... comments above class bodies. These should be proper docstrings ("""...""") so they're accessible programmatically.

7. Test naming
The style guide says: "Name test functions test_<feature>, where <feature> is lowercase and succinct." The current names read like full sentences, e.g. test_median_time_from_infection_to_aids_without_treatment_using_run. Shorter names like test_time_to_aids, test_cd4_decline, test_latent_rel_trans, test_acute_rel_trans are more in line with the convention. Same goes for the class name, but per point 1, the class should go away entirely in favor of plain functions.


Overall the test logic is sound, so these comments are mainly about bringing the style in line with the rest of the codebase. Happy to pair on any of this if helpful!

ckirkman-IDM:
@robynstuart 
Updated code in vein of prior PR (using comments above).

The acute > latent rel_trans test is intended to fail for now until the double stepping bug is fixed in v15 (it passes in manual check with your one-line dt fix put in)

Apparently a starsim fix has been made to address this issue, as a local-update of starsim also passes this test.

robynstuart (approved):
Looks good, pls go ahead and merge! Thanks @ckirkman-IDM

PR #303, opened 2026-02-25
Test ensure HIV AIDS rel_trans > latent rel_trans (Fix #227)

### LINKED ISSUE #227: Test late-stage HIV has higher transmission than chronic phase
Sub-issue of #194.

ckirkman-IDM:
Fix #227 

Tests verify rel_trans falling >= latent (always) and (eventually) falling > latent.

We can add an official 'aids' test once we have an easy handle to grab onto to filter agents in aids stage. (always falling > latent).


robynstuart (changes requested):
Looks good! Butne thing to flag: the elevated `rel_trans` in the falling stage isn't actually applied to all `falling` agents, it's applied to agents with CD4 < 200 (i.e., AIDS). So an agent can be in the `falling` state with `rel_trans == 1` before their CD4 drops below 200. You can see this at [hiv.py L421](https://github.com/starsimhub/stisim/blob/v15/stisim/diseases/hiv.py#L421).

This is actually my bad because we should have had an `aids` property on the HIV module already. I just added one to v15: `hiv.aids` returns `self.cd4 < 200`. Would be great to update the test to use `hiv.aids` instead of `hiv.falling`, which would make it more precise about what's actually being tested.

ckirkman-IDM:
> Looks good! Butne thing to flag: the elevated `rel_trans` in the falling stage isn't actually applied to all `falling` agents, it's applied to agents with CD4 < 200 (i.e., AIDS). So an agent can be in the `falling` state with `rel_trans == 1` before their CD4 drops below 200. You can see this at [hiv.py L421](https://github.com/starsimhub/stisim/blob/v15/stisim/diseases/hiv.py#L421).
> 
> This is actually my bad because we should have had an `aids` property on the HIV module already. I just added one to v15: `hiv.aids` returns `self.cd4 < 200`. Would be great to update the test to use `hiv.aids` instead of `hiv.falling`, which would make it more precise about what's actually being tested.

Will do momentarily, I have merged in v15 to this branch.

ckirkman-IDM:
Updated to use new hiv.aids property in the test.

PR #304, opened 2026-02-25
Test to ensure no sexual transmissions occur in the absence of a sexual network (Fix #238)

### LINKED ISSUE #238: Test that removing sexual networks eliminates transmission
Sub-issue of #194.

ckirkman-IDM:
Now tracking sexual transmissions only (when sexual network removed). Should avoid all other types of infections (e.g. mtct).

#238


robynstuart (commented):
Thanks @ckirkman-IDM! This is definitely a scientifically valid approach, but I think we'd get the same answer from a 3 line version: 
```
def test_no_sexual_transmission_without_network():
    sim = build_testing_sim(sexual_network=None, n_agents=100, duration=1)
    sim.run()
    assert sim.diseases.hiv.new_transmissions_sex.sum() == 0
```
i.e. I don't know if we need the analyzer since the thing we want to test is already stored in the HIV module results. Thoughts?

ckirkman-IDM:
> sim.diseases.hiv.new_transmissions_sex.sum()

I am not quite sure that works. As-is (little details! :)

Agents could die and have their transmissions not counted. This can be partially-fixed/ensured it isn't happening by turning off natural death AND HIV death). 

However, your approach also only checks new_transmissions_sex at the last timestep of the simulation, not throughout the simulation. Running my analyzer in the debugger shows that the value is recorded per-timestep, and dynamically goes up and down per uid). To be truly accurate here requires summing up every timestep, not just end-run checking (unless there is another summary value that can be used on the hiv module).

image

robynstuart:
Ok gotcha. For now, let's merge this. Then later, let's consider if we want to (a) add a new_infections_sex result to the base SEIS class in sti.py (it's a one-liner in set_outcomes), and (b) use ss.infection_log in the longer term once starsim records the network per transmission event (see https://github.com/starsimhub/starsim/issues/1232)

PR #306, opened 2026-02-26
Test HIV doubling (low enough) beta (maternal or sexual) roughly doubles related transmissions

### LINKED ISSUE #237: Test that doubling beta approximately doubles incidence (at low prevalence)
Sub-issue of #194.

ckirkman-IDM:
Discussing with Adam led to getting these tests to pass. The key is low enough _transmission_ (per coital act/mother-to-child transmission step) not low _prevalence_ , to avoid transmission saturation (where an agent is nearly guaranteed to infect their partner/child no matter if you double/halve beta)


tests/test_hiv_natural_history_verification.py
```diff
@@ -126,6 +130,97 @@ def test_aids_transmission_is_higher_than_latent():
...
+    assert hiv_transmissions_test > 0, f"Cannot assess effect of beta, no {mode} HIV transmissions occurred in test."
+
+    # check transmission ratio, expected to be approximately increased by "multiplier" if base beta is low enough
+    expected_ratio = multiplier
+    delta = result_tolerance * expected_ratio
+    test_ratio = hiv_transmissions_test / hiv_transmissions_baseline
+    msg = f"{mode} transmission ratio: {test_ratio} not within delta: {delta} of expected: {expected_ratio}"
+    assert test_ratio == pytest.approx(expected=expected_ratio, rel=result_tolerance), msg
+
+    if verbose:
```
robynstuart, line 189: verbose not defined?

tests/test_hiv_natural_history_verification.py
```diff
@@ -126,6 +130,97 @@ def test_aids_transmission_is_higher_than_latent():
...
+
+
+def _run_beta_test(baseline_m2f, baseline_m2c, mode: str, multiplier=2, result_tolerance=0.1,
+                   n_agents=50000, duration=1, init_prev=0.05, fertility=10):
+    if mode == 'sexual':
+        analyzer = SexualTransmissionCountTracker()
+    elif mode == 'mtc':
+        analyzer = MTCTransmissionCountTracker()
+    else:
+        raise Exception(f'Unknown beta test mode: {mode}')
```
robynstuart: raise ValueError?

robynstuart (commented):
@ckirkman-IDM my main worry with this test is whether it's the kind of thing that's break a lot. Did you have to tune the parameters to get things to work? Do we expect this to break a lot in CI with stochastic variance? Should we bump to 20-25% tolerance?

tests/test_hiv_natural_history_verification.py
ckirkman-IDM: Dreaded global variable! Followed pattern that allows use of it in \_\_main__, applying it here, too.

tests/test_hiv_natural_history_verification.py
ckirkman-IDM: I don't normally worry too much about exception types from test code itself (from expected exceptions in calling real code, YES!) But easy to update.

ckirkman-IDM:
> @ckirkman-IDM my main worry with this test is whether it's the kind of thing that's break a lot. Did you have to tune the parameters to get things to work? Do we expect this to break a lot in CI with stochastic variance? Should we bump to 20-25% tolerance?

I tuned the test size and error together. However, I am now finding the 'rand_seed' sim parameter and will test it out. Update: it does fail after a few random seeds. I will play with test size/tolerance a bit to get it to pass much more reliably.

I have upped the tolerance to 16% (still 3-4% above variance I noticed in 40 random seeds, both tests.) Increased population counts to decrease variance, too.

Issue #241, opened 2026-02-18
Test that ART reduces HIV mortality


robynstuart:
Sub-issue of #195.

ckirkman-IDM:
Test design proposal:

(~ pop=10)
ART coverage 50% of a 100% infected population, no pregnancy, no (natural) death for 20 years
Assert: Zero no-ART agents alive at end
Assert: Non-zero number of ART agents alive at end

robynstuart:
@ckirkman-IDM note that this is partially addressed in PR #320 (`test_hiv_interventions.py`). The `test_art_reduces_mortality` test runs two sims (with and without ART) and compares cumulative deaths. The assertion is currently commented out pending validation with larger populations.

The proposed test design (100% infected, 50% ART, no pregnancy/death, assert all non-ART die) is a good additional test, so let's proceed with your proposed design. With the recent ART refactor you can do

```python
test = sti.HIVTest(test_prob_data=1.0, dt_scale=False)
art = sti.ART(coverage=0.5, art_initiation=1)
```

The key consideration is that HIV mortality acts through CD4 decline, so the sim needs to run long enough for untreated agents' CD4 to reach zero. With `dur=20` this should work.

Issue #314, opened 2026-03-10
HIV interventions are undocumented - experience in using interventions for the first time


ckirkman-IDM:
Using HIV interventions accurately (or easily) is impossible without outside documentation and/or deep reading of code along with repeated debugger stops to learn what is going on.

In trying to make a simple test setup where:

- All agents are HIV+
- All agents are on ART (initiate immediately)

... I ran across numerous issues in trying to do so. The issues boil down to:
- There is no documentation in-code for the intervention classes
- many of the parameters are hard to find (what are they? what can I pass in? where are they used? Are they used in more than one place?)
- no parameter type hints (what can I even pass in? What does e.g. coverage data look like, and do I even need it?)
- parameter names that overlap in potential meaning or are misleading with no documentation. Short names that require interpretation, but > 1 interpretation is possible:
  - e.g. init_prob : Initial probability (like an initial distribution of ART)? (first guess, lead down one rabbit hole) Actual: initiation probability (chance to initiate ART)
  - e.g. ART future_coverage: had to use this for HISTORICAL coverage (starting), not just future). And can I add > 1 timing?
- what is the processing flow of an intervention? How does step(uids) work, where do the uids come from? ...

I eventually figured out that I need to diagnose the agents before I could give them ART, which is not documented. But then I found that not all agents were testing, despite:

```
HIVTest(test_prob_data=1)  # means all agents test, supposedly
```

... but I had to do this (undocumented: test_prob_data is PER YEAR, not per dt, so only 1/12 of agents were testing immediately):

```
HIVTest(test_prob_data=12)  # all agents test immediately with a monthly dt
```

This eventually worked to diagnose all agents on step 0 and then put them all on ART:
```
test_intervention = HIVTest(test_prob_data=12)
initial_art_intervention = ART(init_prob=1.0, future_coverage={'year': 1900, 'prop': 1.0})

```

robynstuart:
@ckirkman-IDM thank you for raising this, it's definitely a high-priority issue. I've added a few sub issues and will prioritize these over the next sprint

robynstuart:
Thanks again for the detailed writeup @ckirkman-IDM, this was SUPER valuable for identifying the pain points! I've opened  #320 to address these. Here's a summary:

Misleading parameter names:
- init_prob → renamed to art_initiation (with deprecation shim for old name)
- future_coverage → deprecated entirely. Use the coverage parameter with extended data instead.
- coverage_data → renamed to coverage (with deprecation shim)

[...]

Issue #313, opened 2026-03-10
HIV ART intervention should skip unborn agent rel_sus update if pregnancy is not included in the sim


ckirkman-IDM:
The following throws an error due to lack of pregnancy attribute if:

```
sim = sti.Sim(demographics=[stuff, no_pregnancy_included])
```

https://github.com/starsimhub/stisim/blob/b446543ea365be23a8f54152b2d90b55f114ea92/stisim/interventions/hiv_interventions.py#L128-L132

robynstuart:
Fixed on v15

Issue #126, opened 2025-05-20
Make ART class more flexible


cliffckerr:
Currently expects ART coverage to be supplied as a dataframe `coverage_data` with index as time and a single column of `n_art` or `p_art`. Would be good to have a bit more flexibility of inputs, e.g. a 2-column dataframe with `years` and `p_art` (or `coverage`), a dict of arrays, etc.

robynstuart:
Potential uses - simple:
```
art = sti.ART(coverage_data=pd.DataFrame(index=np.arange(2000, 2021), data={'p_art': np.linspace(0, 0.9, 21)}))
art = sti.ART(coverage_data=pd.DataFrame(index=np.arange(2000, 2021), data={'n_art': np.linspace(0, 1e6, 21)}))
art = sti.ART(coverage_data=0.9)  # Maintain constant coverage
art = sti.ART(coverage_data={'value':[0, 0.9], 'year':[2000,2020]}  # Should interpolate
```

Also need to plan for more complex use cases where we may have coverage data by age cohort and sex, e.g. with this Eswatini file:

[SWAZILAND_calibration_nationalARTprevalence_2021 copy.csv](https://github.com/user-attachments/files/25910563/SWAZILAND_calibration_nationalARTprevalence_2021.copy.csv)

Additional tasks:
- [ ] Raise a warning if a user tries to add ART without testing
- [ ] Rename init_prob to somehting more descriptive
- [ ] Consider renaming coverage_data
- [ ] Remove the distinction between `future_coverage` and `coverage_data`, it doesn't make sense since future_coverage might be in the past.

PR #320, opened 2026-03-11
Refactor HIV interventions: flexible coverage, documentation, tests

robynstuart:
## Summary

- Refactor ART and VMMC coverage input handling (#126): accept scalar, dict, DataFrame, or age/sex-stratified DataFrame via a unified `coverage` parameter
- Document the HIV testing → diagnosis → ART pipeline (#314): comprehensive docstrings, user guide section with flow diagram, updated tutorial
- Add `art_coverage` analyzer for tracking ART coverage by age/sex with a built-in `plot()` method
- Guard pregnancy/maternalnet access in interventions (#313, merged separately)
- Add `custom` module support to STIsim Sim constructor (merged separately)
- New `test_hiv_interventions.py` with 11 tests covering coverage formats, backward compat, functional effects, and scientific validation skeletons

## Key changes

**ART class:**
- `coverage` replaces `coverage_data` (deprecated with FutureWarning)
- `art_initiation` replaces `init_prob` (deprecated with FutureWarning)
- `future_coverage` deprecated
- No coverage = no force-fitting, just treat all who initiate
- Warns if no HIVTest intervention found
- Supports age/sex-stratified coverage DataFrames

**VMMC class:** Same coverage refactor pattern

**Documentation:**
- `test_prob_data` clearly documented as an annual rate (dt_scale=True) with `dt_scale=False` for per-timestep probability
- User guide: full HIV pipeline section with flow diagram, parameter tables, coverage format examples
- Tutorial: updated to use `hivsim.demo`, new ART coverage formats section, `art_coverage` analyzer demo

## Test plan

- [x] All 11 new tests in `test_hiv_interventions.py` pass
- [x] All 6 existing tests in `test_sim.py` pass (with expected deprecation warnings)
- [x] Backward compat verified with `anc_sti_screening` repo
- [ ] Review tutorial notebook renders correctly

### DIFF
--- a/docs/user_guide/interventions/interventions.md
+++ b/docs/user_guide/interventions/interventions.md
@@ -179,7 +179,7 @@ art = sti.ART(coverage=stratified_df)
 
 | Parameter | Default | Description |
 |-----------|---------|-------------|
-| `coverage` | None (no target; treat all who initiate) | Coverage target: scalar, dict, DataFrame, or stratified DataFrame. Dict values are linearly interpolated between years. DataFrame requires index=years and a column named `n_art` (absolute numbers) or `p_art` (proportion of infected). Stratified DataFrames need columns Year, Gender/Sex, AgeBin `[lo,hi)`, and a numeric value column. |
+| `coverage` | None (no target; treat all who initiate) | Coverage target: scalar, dict, DataFrame, or stratified DataFrame. Dict values are linearly interpolated between years. DataFrame requires index=years and a column named `n_art` (absolute numbers) or `p_art` (proportion of infected). Stratified DataFrames need columns Year, Gender/Sex, AgeBin `[lo,hi)`, and a numeric value column. Gender accepts: 0/f/female (female), 1/m/male (male). |
 | `art_initiation` | bernoulli(0.9) | Probability a newly diagnosed person initiates ART. Set to 1 to treat all diagnosed. |
 
 ### VMMC
--- a/docs/user_guide/interventions/interventions.md
+++ b/docs/user_guide/interventions/interventions.md
@@ -157,8 +157,11 @@ fsw_test = sti.HIVTest(
 ART accepts coverage in multiple formats:
 
 ```python
-# No coverage target — treat all who are diagnosed
+# No coverage target — 90% of newly diagnosed initiate ART (default)
 art = sti.ART()
+
+# Treat ALL diagnosed, no coverage constraint
+art = sti.ART(art_initiation=1)
 
 # Constant proportion of infected
 art = sti.ART(coverage=0.8)
--- a/docs/user_guide/interventions/interventions.md
+++ b/docs/user_guide/interventions/interventions.md
@@ -28,7 +28,7 @@ The base testing class. Controls who gets tested, how often, and with what diagn
 ```python
 test = sti.STITest(
     product=my_diagnostic,      # Diagnostic product (required)
-    test_prob_data=0.1,         # Per-timestep probability of testing
+    test_prob_data=0.1,         # Annual testing rate (10% per year)
     eligibility=lambda sim: sim.people.female,  # Who is eligible
     start=2015,                 # When testing begins
 )
@@ -39,10 +39,10 @@ Key parameters:
 | Parameter | Default | Description |
 |-----------|---------|-------------|
 | `product` | (required) | Diagnostic product (e.g., `STIDx`, `HIVDx`) |
-| `test_prob_data` | 1.0 | Testing probability -- scalar, array, or time-varying |
+| `test_prob_data` | 1.0 | Annual testing rate (scalar or array over years). Automatically scaled by dt when `dt_scale=True`. |
 | `eligibility` | all agents | Function `f(sim) -> UIDs` defining who can be tested |
 | `start` / `stop` | sim start/end | Active period for the intervention |
-| `dt_scale` | True | Scale probability by timestep length |
+| `dt_scale` | True | If True, interpret `test_prob_data` as an annual rate and multiply by dt. Set to False to use as a per-timestep probability. |
 
 ### HIVTest
 
@@ -127,7 +127,7 @@ hiv.ti_diagnosed = ti         │
 **Key points:**
 
 - Agents must be **diagnosed** before they can go on ART. If you add ART without HIVTest, a warning is raised and no agents will be treated.
-- `test_prob_data` is a **per-year rate**, automatically scaled by dt. With monthly timesteps (dt=1/12), `test_prob_data=0.1` means ~0.83% per month, not 10% per month. To test everyone in one timestep, use `test_prob_data=1/dt` (e.g. 12 for monthly).
+- `test_prob_data` is an **annual testing rate** by default (`dt_scale=True`). With monthly timesteps (dt=1/12), `test_prob_data=0.1` means ~0.83% per month, or ~10% per year. To use a per-timestep probability directly, set `dt_scale=False`.
 - ART `art_initiation` (default 0.9) controls what fraction of newly diagnosed agents are willing to start treatment.
 - If `coverage` is provided, ART force-fits the number on treatment to match the target by adding/removing agents based on CD4 and care-seeking.
 - If no `coverage` is provided (the default), ART simply treats everyone who initiates, with no capacity constraint.
@@ -148,7 +148,7 @@ fsw_test = sti.HIVTest(
 
 | Parameter | Default | Description |
 |-----------|---------|-------------|
-| `test_prob_data` | 1.0 | Per-year testing probability (scaled by dt) |
+| `test_prob_data` | 1.0 | Annual testing rate (scaled by dt when `dt_scale=True`) |
 | `eligibility` | undiagnosed | Function `f(sim) -> BoolArr` defining who can be tested |
 | `start` | sim start | Year testing begins |
 
--- a/docs/user_guide/interventions/interventions.md
+++ b/docs/user_guide/interventions/interventions.md
@@ -38,11 +38,11 @@ Key parameters:
 
 | Parameter | Default | Description |
 |-----------|---------|-------------|
-| `product` | (required) | Diagnostic product (e.g., `STIDx`, `HIVDx`) |
-| `test_prob_data` | 1.0 | Annual testing rate (scalar or array over years). Automatically scaled by dt when `dt_scale=True`. |
-| `eligibility` | all agents | Function `f(sim) -> UIDs` defining who can be tested |
-| `start` / `stop` | sim start/end | Active period for the intervention |
-| `dt_scale` | True | If True, interpret `test_prob_data` as an annual rate and multiply by dt. Set to False to use as a per-timestep probability. |
+| `product` | (required for STITest; HIVTest auto-creates a perfect HIVDx) | Diagnostic product (e.g., `STIDx`, `HIVDx`) |
+| `test_prob_data` | 1.0 | Annual testing probability (scalar or array over years). Converted to per-timestep probability via `ss.probperyear` when `dt_scale=True`. |
+| `eligibility` | all agents (HIVTest: undiagnosed only) | Function `f(sim) -> BoolArr` defining who can be tested, e.g. `lambda sim: sim.people.female & (sim.people.age >= 15)` |
+| `start` / `stop` | first/last sim year | Calendar year (inclusive) when the intervention activates/deactivates |
+| `dt_scale` | True | If True (default), interpret `test_prob_data` as an annual probability. Set to False to interpret as a per-timestep probability. |
 
 ### HIVTest
 
@@ -126,10 +126,11 @@ hiv.ti_diagnosed = ti         │
 
 **Key points:**
 
+- **Intervention order matters**: interventions run in the order they appear in the list. HIVTest must come before ART so that agents diagnosed this timestep can initiate ART in the same step.
 - Agents must be **diagnosed** before they can go on ART. If you add ART without HIVTest, a warning is raised and no agents will be treated.
-- `test_prob_data` is an **annual testing rate** by default (`dt_scale=True`). With monthly timesteps (dt=1/12), `test_prob_data=0.1` means ~0.83% per month, or ~10% per year. To use a per-timestep probability directly, set `dt_scale=False`.
+- `test_prob_data` is an **annual testing probability** by default (`dt_scale=True`). It is converted to a per-timestep probability via `ss.probperyear`. With monthly timesteps, `test_prob_data=0.1` means ~0.88% per month, correctly recovering ~10% per year. To use a per-timestep probability directly, set `dt_scale=False`.
 - ART `art_initiation` (default 0.9) controls what fraction of newly diagnosed agents are willing to start treatment.
-- If `coverage` is provided, ART force-fits the number on treatment to match the target by adding/removing agents based on CD4 and care-seeking.
+- If `coverage` is provided, ART force-fits the number on treatment to match the target each timestep. Agents are added or removed immediately (not gradually) to hit the target, prioritized by CD4 count and care-seeking propensity.
 - If no `coverage` is provided (the default), ART simply treats everyone who initiates, with no capacity constraint.
 
 ### HIVTest
@@ -148,7 +149,7 @@ fsw_test = sti.HIVTest(
 
 | Parameter | Default | Description |
 |-----------|---------|-------------|
-| `test_prob_data` | 1.0 | Annual testing rate (scaled by dt when `dt_scale=True`) |
+| `test_prob_data` | 1.0 | Annual testing probability (converted via `ss.probperyear` when `dt_scale=True`) |
 | `eligibility` | undiagnosed | Function `f(sim) -> BoolArr` defining who can be tested |
 | `start` | sim start | Year testing begins |
 
@@ -178,12 +179,12 @@ art = sti.ART(coverage=stratified_df)
 
 | Parameter | Default | Description |
 |-----------|---------|-------------|
-| `coverage` | None | Coverage target: scalar, dict, DataFrame, or stratified DataFrame |
-| `art_initiation` | bernoulli(0.9) | Probability a newly diagnosed person initiates ART |
+| `coverage` | None (no target; treat all who initiate) | Coverage target: scalar, dict, DataFrame, or stratified DataFrame. Dict values are linearly interpolated between years. DataFrame requires index=years and a column named `n_art` (absolute numbers) or `p_art` (proportion of infected). Stratified DataFrames need columns Year, Gender/Sex, AgeBin `[lo,hi)`, and a numeric value column. |
+| `art_initiation` | bernoulli(0.9) | Probability a newly diagnosed person initiates ART. Set to 1 to treat all diagnosed. |
 
 ### VMMC
 
-Voluntary medical male circumcision. Reduces male susceptibility to HIV by 60%.
+Voluntary medical male circumcision. Targets all males by default; reduces susceptibility to HIV acquisition by 60%.
 
 ```python
 vmmc = sti.VMMC(coverage=0.3)
@@ -192,12 +193,13 @@ vmmc = sti.VMMC(coverage={'year': [2010, 2025], 'value': [0, 0.4]})
 
 | Parameter | Default | Description |
 |-----------|---------|-------------|
-| `coverage` | None | Coverage target (same formats as ART) |
+| `coverage` | None (does nothing without coverage data) | Coverage target (same formats as ART). Required for VMMC to have any effect. |
 | `eff_circ` | 0.6 | Efficacy (60% reduction in HIV acquisition risk) |
+| `eligibility` | all males | Optional function to further restrict who is eligible |
 
 ### PrEP
 
-Pre-exposure prophylaxis targeting FSWs. Reduces susceptibility by 80%.
+Pre-exposure prophylaxis. By default targets HIV-negative FSWs (female sex workers) who are not already on PrEP. Use the `eligibility` parameter to target a different population.
 
 ```python
 prep = sti.Prep(coverage=[0, 0.5], years=[2020, 2025])
@@ -205,9 +207,10 @@ prep = sti.Prep(coverage=[0, 0.5], years=[2020, 2025])
 
 | Parameter | Default | Description |
 |-----------|---------|-------------|
-| `coverage` | ramps to 80% | Coverage at each year |
-| `years` | [2004, 2005, 2015, 2025] | Corresponding years |
-| `eff_prep` | 0.8 | Efficacy (80% reduction) |
+| `coverage` | [0, 0.01, 0.5, 0.8] | Coverage values at each year (linearly interpolated between years) |
+| `years` | [2004, 2005, 2015, 2025] | Calendar years corresponding to coverage values |
+| `eff_prep` | 0.8 | Efficacy (80% reduction in HIV acquisition risk) |
+| `eligibility` | HIV-negative FSWs not on PrEP | Optional function to override default targeting |
 
 ## Combining interventions
 
--- a/docs/user_guide/interventions/interventions.md
+++ b/docs/user_guide/interventions/interventions.md
@@ -100,33 +100,111 @@ Treatment outcomes are tracked as successful (infection cleared), failed (infect
 | `sti.SyphTx` | Syphilis | Treats pregnant mothers and fetuses; stage-specific |
 | `sti.treat_BV` | BV | CST-based treatment with durable post-treatment effects |
 
-## HIV-specific interventions
+## HIV treatment pipeline
+
+HIV treatment in STIsim follows a specific pipeline. Understanding this flow is essential for setting up HIV interventions correctly:
+
+```
+HIVTest                          ART
+  │                               │
+  │ test eligible agents          │ check newly diagnosed
+  │ (per-year probability)        │ (ti_diagnosed == this step)
+  │                               │
+  ▼                               ▼
+hiv.diagnosed = True    ──►   art_initiation filter (90% default)
+hiv.ti_diagnosed = ti         │
+                              ▼
+                         If coverage specified:
+                           prioritize by CD4, match target
+                         If no coverage:
+                           treat all who initiate
+                              │
+                              ▼
+                         hiv.on_art = True
+                         (reduces transmissibility, reconstitutes CD4)
+```
+
+**Key points:**
+
+- Agents must be **diagnosed** before they can go on ART. If you add ART without HIVTest, a warning is raised and no agents will be treated.
+- `test_prob_data` is a **per-year rate**, automatically scaled by dt. With monthly timesteps (dt=1/12), `test_prob_data=0.1` means ~0.83% per month, not 10% per month. To test everyone in one timestep, use `test_prob_data=1/dt` (e.g. 12 for monthly).
+- ART `art_initiation` (default 0.9) controls what fraction of newly diagnosed agents are willing to start treatment.
+- If `coverage` is provided, ART force-fits the number on treatment to match the target by adding/removing agents based on CD4 and care-seeking.
+- If no `coverage` is provided (the default), ART simply treats everyone who initiates, with no capacity constraint.
+
+### HIVTest
+
+```python
+# Test 20% of undiagnosed agents per year
+hiv_test = sti.HIVTest(test_prob_data=0.2, start=2000, name='hiv_test')
+
+# Higher rate for FSWs
+fsw_test = sti.HIVTest(
+    test_prob_data=0.5,
+    name='fsw_test',
+    eligibility=lambda sim: sim.networks.structuredsexual.fsw & ~sim.diseases.hiv.diagnosed,
+)
+```
+
+| Parameter | Default | Description |
+|-----------|---------|-------------|
+| `test_prob_data` | 1.0 | Per-year testing probability (scaled by dt) |
+| `eligibility` | undiagnosed | Function `f(sim) -> BoolArr` defining who can be tested |
+| `start` | sim start | Year testing begins |
 
 ### ART
 
-Antiretroviral therapy with CD4-based prioritization and coverage targets.
+ART accepts coverage in multiple formats:
 
 ```python
-art = sti.ART(coverage_data=art_df)  # DataFrame with year index and n_art or p_art column
+# No coverage target — treat all who are diagnosed
+art = sti.ART()
+
+# Constant proportion of infected
+art = sti.ART(coverage=0.8)
+
+# Time-varying proportion
+art = sti.ART(coverage={'year': [2000, 2010, 2025], 'value': [0, 0.5, 0.9]})
+
+# From a DataFrame (absolute numbers)
+art = sti.ART(coverage=pd.read_csv('art_data.csv').set_index('year'))
+
+# Age/sex stratified (columns: Year, Gender/Sex, AgeBin, + value)
+art = sti.ART(coverage=stratified_df)
 ```
 
-ART reduces transmissibility by 96% (default) and reconstitutes CD4 counts. Coverage can be specified as absolute numbers or proportions.
+| Parameter | Default | Description |
+|-----------|---------|-------------|
+| `coverage` | None | Coverage target: scalar, dict, DataFrame, or stratified DataFrame |
+| `art_initiation` | bernoulli(0.9) | Probability a newly diagnosed person initiates ART |
 
 ### VMMC
 
-Voluntary medical male circumcision. Reduces male susceptibility to HIV by 60% (default).
+Voluntary medical male circumcision. Reduces male susceptibility to HIV by 60%.
 
 ```python
-vmmc = sti.VMMC(coverage_data=vmmc_df)
+vmmc = sti.VMMC(coverage=0.3)
+vmmc = sti.VMMC(coverage={'year': [2010, 2025], 'value': [0, 0.4]})
 ```
+
+| Parameter | Default | Description |
+|-----------|---------|-------------|
+| `coverage` | None | Coverage target (same formats as ART) |
+| `eff_circ` | 0.6 | Efficacy (60% reduction in HIV acquisition risk) |
 
 ### PrEP
 
-Pre-exposure prophylaxis targeting FSWs. Reduces susceptibility by 80% (default).
+Pre-exposure prophylaxis targeting FSWs. Reduces susceptibility by 80%.
 
 ```python
 prep = sti.Prep(coverage=[0, 0.5], years=[2020, 2025])
 ```
+
+| Parameter | Default | Description |
+|-----------|---------|-------------|
+| `coverage` | ramps to 80% | Coverage at each year |
+| `years` | [2004, 2005, 2015, 2025] | Corresponding years |
+| `eff_prep` | 0.8 | Efficacy (80% reduction) |
 
 ## Combining interventions
 
--- a/stisim/analyzers.py
+++ b/stisim/analyzers.py
@@ -639,3 +639,52 @@ class art_coverage(ss.Analyzer):
 
         return
 
+    def plot(self, by_age=True):
+        """
+        Plot ART coverage over time.
+
+        Creates a 2-panel figure: aggregate coverage (left) and by age/sex (right).
+        If by_age=False, only plots aggregate.
+
+        Example::
+
+            sim.run()
+            sim.analyzers.art_coverage.plot()
+        """
+        yearvec = self.sim.t.yearvec
+        sex_colors = {'f': '#d46e9c', 'm': '#4a90d9'}
+        sex_labels = {'f': 'Women', 'm': 'Men'}
+
+        if by_age and len(self.age_bins) > 2:
+            fig, axes = pl.subplots(1, 3, figsize=(16, 5))
+        else:
+            fig, axes = pl.subplots(1, 1, figsize=(6, 5))
+            axes = [axes]
+
+        # Panel 1: aggregate
+        ax = axes[0]
+        ax.plot(yearvec, self.results['p_art'],   color='k',              linewidth=2, label='Overall')
+        ax.plot(yearvec, self.results['p_art_f'], color=sex_colors['f'], linewidth=1.5, label='Women')
+        ax.plot(yearvec, self.results['p_art_m'], color=sex_colors['m'], linewidth=1.5, label='Men')
+        ax.set_xlabel('Year')
+        ax.set_ylabel('ART coverage (proportion of infected)')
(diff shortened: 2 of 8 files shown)


stisim/interventions/base_interventions.py
```diff
@@ -58,7 +58,40 @@ def administer(self, sim, uids):
...
+        product (ss.Product):  diagnostic product that determines test outcomes
+        test_prob_data:        annual testing rate (if dt_scale=True, the default)
+                               or per-timestep probability (if dt_scale=False);
+                               scalar or array over years
+        years (array):         years corresponding to test_prob_data if array
+        start/stop (float):    active period for the intervention
+        eligibility (func):    function f(sim) -> BoolArr or UIDs defining who can be tested
+        rel_test (float):      relative testing probability multiplier (default 1)
+        dt_scale (bool):       if True (default), interpret test_prob_data as an annual
+                               rate and multiply by dt to get per-timestep probability.
```
ckirkman-IDM: I would simplify the True part of the description and mirror the True/False language to more clearly highlight the difference between True/False like so:


> if True (default), interpret test_prob_data as an annual rate. Set to False to interpret test_prob_data as a per-timestep rate.

stisim/interventions/base_interventions.py
```diff
@@ -58,7 +58,40 @@ def administer(self, sim, uids):
...
+                               Set to False to use test_prob_data as-is per timestep.
+
+    States set on agents:
+        tested (bool):      ever tested
+        diagnosed (bool):   ever diagnosed positive
+        ti_tested (float):  timestep of last test
+        ti_positive (float): timestep of last positive result
```
ckirkman-IDM, line 108: There are more states below, is there a reason why they aren't being documented, too? Lines 125-134 .

stisim/interventions/base_interventions.py
```diff
@@ -58,7 +58,40 @@ def administer(self, sim, uids):
...
 class STITest(ss.Intervention):
     """
-    Base class for STI testing
+    Base class for STI testing.
+
+    Controls who gets tested, how often, and with what diagnostic product.
+    Each timestep, eligible agents are tested with a probability derived from
+    test_prob_data. By default (dt_scale=True), test_prob_data is interpreted
+    as an **annual testing rate** and automatically scaled by dt to get a
+    per-timestep probability.
```
ckirkman-IDM: Just a detail, perhaps a follow-on issue for another time.

By annual rate, we are currently meaning: test_prob_data / n_timesteps_per_year .

But in reality, that is not right. For example:
Actual data: 24% test per year (assume no double testers)

This is what we currently do to determine per-dt testing.
24% test per year with dt=month -> 24/12 = 2% test per month (98% don't test per month)

But if you then ask, "How many test in a year?" (assume no double testers):
everyone - (remaining non-testers)**(n_testing_checks) = n_testers
1 - (1 - 0.02)**12 = 0.2153 = 21.53%
We do not recover the original stated 24%/year

stisim/interventions/base_interventions.py
```diff
@@ -58,7 +58,40 @@ def administer(self, sim, uids):
...
+
+    If you need to specify a per-timestep probability directly (not an annual
+    rate), set dt_scale=False:
+        - test_prob_data=0.5, dt_scale=False → 50% chance per timestep
+
+    Args:
+        product (ss.Product):  diagnostic product that determines test outcomes
+        test_prob_data:        annual testing rate (if dt_scale=True, the default)
+                               or per-timestep probability (if dt_scale=False);
+                               scalar or array over years
```
ckirkman-IDM: We need to document here by example and/or types, or redirect to where such information can be found regarding what can be passed as a test_prob_data, otherwise a user has to pass in stuff at random(ish) and see what happens (if they have a diagnostic to help them).

stisim/interventions/base_interventions.py
```diff
@@ -58,7 +58,40 @@ def administer(self, sim, uids):
...
+        years (array):         years corresponding to test_prob_data if array
+        start/stop (float):    active period for the intervention
```
ckirkman-IDM: Timesteps? Years? Calendar years? Years from beginning of simulation? Units, please. Also, are the start/stop inclusive? For example, 2040-2050, is this 2040 < time < 2050, 2040 < time < (2050 +1), ... ?

stisim/interventions/base_interventions.py
```diff
@@ -58,7 +58,40 @@ def administer(self, sim, uids):
...
+        eligibility (func):    function f(sim) -> BoolArr or UIDs defining who can be tested
```
ckirkman-IDM: I assume this eligibility function is what handles any demographic targeting? e.g. women 15-45 with sex_work_intensity > xyz? If so, it might be nice to include a brief mention to make this a bit more descriptive, perhaps linking/pointing to further docs on what can/can't be done.

Also, if not specified, what is the default behavior?

stisim/interventions/base_interventions.py
```diff
@@ -58,7 +58,40 @@ def administer(self, sim, uids):
...
+    For example, with monthly timesteps (dt=1/12):
+        - test_prob_data=0.1  → ~0.83% per month → ~10% tested per year
+        - test_prob_data=1.0  → ~8.3% per month  → ~100% tested per year
+
+    If you need to specify a per-timestep probability directly (not an annual
+    rate), set dt_scale=False:
+        - test_prob_data=0.5, dt_scale=False → 50% chance per timestep
+
+    Args:
+        product (ss.Product):  diagnostic product that determines test outcomes
```
ckirkman-IDM: What if there is no product? (default behavior)

docs/user_guide/interventions/interventions.md
```diff
@@ -100,34 +100,115 @@ Treatment outcomes are tracked as successful (infection cleared), failed (infect
...
 
 ### VMMC
 
-Voluntary medical male circumcision. Reduces male susceptibility to HIV by 60% (default).
+Voluntary medical male circumcision. Reduces male susceptibility to HIV by 60%.
 
 ```python
-vmmc = sti.VMMC(coverage_data=vmmc_df)
+vmmc = sti.VMMC(coverage=0.3)
+vmmc = sti.VMMC(coverage={'year': [2010, 2025], 'value': [0, 0.4]})
```
ckirkman-IDM, line 191: mirroring a comment in base_interventions.py, what is the default behavior? (as in, who is targeted here). I'm currently trying to get a sense of where default behaviors are noted, what they are, how to find where to override them, ...

docs/user_guide/interventions/interventions.md
```diff
@@ -100,34 +100,115 @@ Treatment outcomes are tracked as successful (infection cleared), failed (infect
...
+                              ▼
+                         If coverage specified:
+                           prioritize by CD4, match target
+                         If no coverage:
+                           treat all who initiate
+                              │
+                              ▼
+                         hiv.on_art = True
+                         (reduces transmissibility, reconstitutes CD4)
+```
```
ckirkman-IDM, line 125: I like this flowchart, it is super helpful. The only detail I can think of is that order of interventions matters. If you want to put agents on ART diagnosed THIS timestep, the diagnosis has to come before the ART. Somewhere, someway, that mental map of ordered interventions needs to be documented/described/ingrained in future users.

docs/user_guide/interventions/interventions.md
```diff
@@ -100,34 +100,115 @@ Treatment outcomes are tracked as successful (infection cleared), failed (infect
...
+
+**Key points:**
+
+- Agents must be **diagnosed** before they can go on ART. If you add ART without HIVTest, a warning is raised and no agents will be treated.
+- `test_prob_data` is an **annual testing rate** by default (`dt_scale=True`). With monthly timesteps (dt=1/12), `test_prob_data=0.1` means ~0.83% per month, or ~10% per year. To use a per-timestep probability directly, set `dt_scale=False`.
+- ART `art_initiation` (default 0.9) controls what fraction of newly diagnosed agents are willing to start treatment.
+- If `coverage` is provided, ART force-fits the number on treatment to match the target by adding/removing agents based on CD4 and care-seeking.
+- If no `coverage` is provided (the default), ART simply treats everyone who initiates, with no capacity constraint.
```
ckirkman-IDM, line 134: Please make sure all default behaviors are recorded here (user doc) and in the class!

docs/user_guide/interventions/interventions.md
```diff
@@ -100,34 +100,115 @@ Treatment outcomes are tracked as successful (infection cleared), failed (infect
...
+                         hiv.on_art = True
+                         (reduces transmissibility, reconstitutes CD4)
+```
+
+**Key points:**
+
+- Agents must be **diagnosed** before they can go on ART. If you add ART without HIVTest, a warning is raised and no agents will be treated.
+- `test_prob_data` is an **annual testing rate** by default (`dt_scale=True`). With monthly timesteps (dt=1/12), `test_prob_data=0.1` means ~0.83% per month, or ~10% per year. To use a per-timestep probability directly, set `dt_scale=False`.
+- ART `art_initiation` (default 0.9) controls what fraction of newly diagnosed agents are willing to start treatment.
+- If `coverage` is provided, ART force-fits the number on treatment to match the target by adding/removing agents based on CD4 and care-seeking.
```
ckirkman-IDM: I understand the adding/removing agents from ART to hit a coverage level, and there may be some super-fine detail here regarding how fast agents are taken off (immediate step jump vs gradual, like in uptake), but some additional doc around this here/in class would be helpful.

docs/user_guide/interventions/interventions.md
```diff
@@ -100,34 +100,115 @@ Treatment outcomes are tracked as successful (infection cleared), failed (infect
...
+    name='fsw_test',
+    eligibility=lambda sim: sim.networks.structuredsexual.fsw & ~sim.diseases.hiv.diagnosed,
+)
+```
+
+| Parameter | Default | Description |
+|-----------|---------|-------------|
+| `test_prob_data` | 1.0 | Annual testing rate (scaled by dt when `dt_scale=True`) |
+| `eligibility` | undiagnosed | Function `f(sim) -> BoolArr` defining who can be tested |
+| `start` | sim start | Year testing begins |
```
ckirkman-IDM, line 154: Example is nice.

docs/user_guide/interventions/interventions.md
```diff
@@ -100,34 +100,115 @@ Treatment outcomes are tracked as successful (infection cleared), failed (infect
...
+art = sti.ART(art_initiation=1)
+
+# Constant proportion of infected
+art = sti.ART(coverage=0.8)
+
+# Time-varying proportion
+art = sti.ART(coverage={'year': [2000, 2010, 2025], 'value': [0, 0.5, 0.9]})
+
+# From a DataFrame (absolute numbers)
+art = sti.ART(coverage=pd.read_csv('art_data.csv').set_index('year'))
```
ckirkman-IDM, line 174: Any dataframe format requirements?

docs/user_guide/interventions/interventions.md
```diff
@@ -100,34 +100,115 @@ Treatment outcomes are tracked as successful (infection cleared), failed (infect
...
 
+| Parameter | Default | Description |
+|-----------|---------|-------------|
+| `coverage` | None | Coverage target (same formats as ART) |
+| `eff_circ` | 0.6 | Efficacy (60% reduction in HIV acquisition risk) |
+
 ### PrEP
 
-Pre-exposure prophylaxis targeting FSWs. Reduces susceptibility by 80% (default).
+Pre-exposure prophylaxis targeting FSWs. Reduces susceptibility by 80%.
```
ckirkman-IDM: Why remove default? What product is assumed? How does a user know 80%?

stisim/analyzers.py
```diff
@@ -563,3 +563,128 @@ def plot(self):
...
+        analyzer.results['p_art_m_25_35']    # Men 25-35 on ART (proportion of infected)
+
+    Args:
+        age_bins (list): age bin edges, e.g. [15, 25, 35, 45, 65]. Default: [15, 25, 35, 45, 65]
+    """
+
+    def __init__(self, age_bins=None, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.name = 'art_coverage'
+        self.age_bins = age_bins or [15, 25, 35, 45, 65]
```
ckirkman-IDM, line 583: Any time I see age bins I ask the same thing. What do the edges mean? The reason is that I have seen users misuse the edges MANY times in the same ways that can cause double or under counting on the edges, e.g.:

15-25, 25-35: 
15 < age < 25 , 25 < age < 35?
15 <= age <= (25+1) , (25 + 1) < age <=+ (35+1)  # note the double counting of float ~25-26 ages
+1 because in real life, if we are 35 and 11.9999 months old we say we are 35 (yes, this makes its way into user code ... and dev code!)

I ASSUME it means: 15 <= age < 25, 25 <= age < 35 , as it often does, but ... there is so much room for misinterpretation on any/all bins! (age, date, ...)

stisim/analyzers.py
```diff
@@ -563,3 +563,128 @@ def plot(self):
...
+        return
+
+    def init_results(self):
+        super().init_results()
+        results = []
+
+        # Aggregate results
+        results.append(ss.Result('n_art',   dtype=int,   label='Total on ART'))
```
ckirkman-IDM, line 591: I like these labels! We should have/use them on "pars"!

stisim/analyzers.py
```diff
@@ -563,3 +563,128 @@ def plot(self):
...
+        self.results['p_art'][ti]   = sc.safedivide(n_art, n_inf)
+        self.results['n_art_f'][ti] = len((on_art & female).uids)
+        self.results['n_art_m'][ti] = len((on_art & male).uids)
+        self.results['p_art_f'][ti] = sc.safedivide(len((on_art & female).uids), len((infected & female).uids))
+        self.results['p_art_m'][ti] = sc.safedivide(len((on_art & male).uids), len((infected & male).uids))
+
+        # Per age bin × sex
+        for i in range(len(self.age_bins) - 1):
+            lo, hi = self.age_bins[i], self.age_bins[i + 1]
+            age_mask = (age >= lo) & (age < hi)
```
ckirkman-IDM, line 633: Answers my age bin question (phew, a sane definition here). But user/dev misuse is still easy/bound to happen.

stisim/analyzers.py
```diff
@@ -563,3 +563,128 @@ def plot(self):
...
+        n_art = len(on_art.uids)
+        n_inf = len(infected.uids)
+        self.results['n_art'][ti]   = n_art
+        self.results['p_art'][ti]   = sc.safedivide(n_art, n_inf)
+        self.results['n_art_f'][ti] = len((on_art & female).uids)
+        self.results['n_art_m'][ti] = len((on_art & male).uids)
+        self.results['p_art_f'][ti] = sc.safedivide(len((on_art & female).uids), len((infected & female).uids))
+        self.results['p_art_m'][ti] = sc.safedivide(len((on_art & male).uids), len((infected & male).uids))
+
+        # Per age bin × sex
```
ckirkman-IDM, line 630: Nice clear logic, I like it. Removing the 'ppl' and 'self.sim.people' prefixes definitely helps readability.

stisim/interventions/hiv_interventions.py
```diff
@@ -294,6 +692,24 @@ def step(self):
...
 
+# %% Utility functions
+def _parse_age_bin(ab_str):
+    """
+    Parse age bin string like "[15,25)" into (lo, hi).
+    Also handles numeric tuples and plain strings like "15-25".
+    """
+    if isinstance(ab_str, (tuple, list)):
+        return float(ab_str[0]), float(ab_str[1])
+    s = str(ab_str).strip('[]() ')
```
ckirkman-IDM, line 744: Technically this would parse:

[15,25) and (15,25] (and any variant) the same, which seems potentially problematic to me. Perhaps, since the bins are used as [x,y), we check for any other bracket-type format and error if found?

stisim/interventions/hiv_interventions.py
```diff
@@ -159,119 +498,178 @@ def prioritize_art(self, sim, n=None, awaiting_art_uids=None):
...
+
+    Reduces HIV susceptibility by eff_prep (default 80%) for eligible FSWs
+    who are HIV-negative and not already on PrEP. Coverage ramps up over
+    time according to the years/coverage parameters.
+
+    Args:
+        coverage (list):  coverage values at each year (default ramps 0→80%)
+        years (list):     corresponding years
+        eff_prep (float): efficacy (default 0.8 = 80% reduction in acquisition)
+        eligibility:      optional custom eligibility function
```
ckirkman-IDM: I see the description says "FSWs", but the inclusion of an eligibility function (for further targeting) is a bit confusing. What if non-FSWs are intended to be targeted? Perhaps we need to (possibly in another issue) change the default eligibility to be FSWs (via function) and update the docs to match this?

stisim/interventions/hiv_interventions.py
```diff
@@ -159,119 +498,178 @@ def prioritize_art(self, sim, n=None, awaiting_art_uids=None):
...
+
+    Example::
+
+        prep = sti.Prep(coverage=[0, 0.5], years=[2020, 2025])
+    """
```
ckirkman-IDM, line 707: Same time question as in other places. Is this a piecewise function of coverage with step jumps, linearly interpolated, how are bounds defined/used, ...

stisim/interventions/hiv_interventions.py
ckirkman-IDM, line 725: Interpolation. Need docs for this.

Do we need the ability (another issue) to turn off interpolation if desired? (for step-jump changes) Or we should document somewhere how to approximate (effectively force, based on dt) non-interpolation at certain timepoints.

stisim/interventions/hiv_interventions.py
```diff
@@ -159,119 +498,178 @@ def prioritize_art(self, sim, n=None, awaiting_art_uids=None):
...
         )
         return
 
+    def _get_n_to_circ(self, eligible_uids):
+        """ Get the target number of circumcisions this timestep """
+        # Legacy future_coverage mode
+        if self._future_coverage is not None and self.t.now('year') >= self._future_coverage['year']:
+            return int(self._future_coverage['prop'] * len(eligible_uids))
+
+        if self.coverage is None:
```
ckirkman-IDM, line 635: While usage of this is correct (checking for None), I would personally prefer a 0 return and remove the None check. Unless there is a real difference between returning None and 0, it would remove the implicit assertion that None and 0 are different (and need to be handled differently) below:

```
if n_to_circ is not None and n_to_circ > 0:
    eligible_uids = (sim.people.male & ~self.circumcised).uids
```

stisim/interventions/hiv_interventions.py
```diff
@@ -159,119 +498,178 @@ def prioritize_art(self, sim, n=None, awaiting_art_uids=None):
...
+        if isinstance(self.coverage, dict):
+            return self._get_n_stratified(eligible_uids)
+
+        # Aggregate coverage
+        if self.coverage_format == 'n':
+            return int(self.coverage[self.ti] / self.sim.pars.pop_scale)
+        else:
+            return int(self.coverage[self.ti] * len(eligible_uids))
+
+    def _get_n_stratified(self, eligible_uids):
```
ckirkman-IDM, line 648: I like this format ... I think (from skimming :D)!

Maybe it should be generalized from VMMC and included somewhere for use ANYWHERE (other interventions) (not this issue/PR, for another).

stisim/interventions/hiv_interventions.py
```diff
@@ -159,119 +498,178 @@ def prioritize_art(self, sim, n=None, awaiting_art_uids=None):
...
+    Voluntary medical male circumcision.
+
+    Reduces male susceptibility to HIV acquisition by eff_circ (default 60%).
+    Unlike ART, VMMC does not require diagnosis — it circumcises males up to a
+    coverage target, prioritized by willingness (a random per-agent score).
+
+    If no coverage is specified, VMMC does nothing. Coverage must be provided
+    explicitly via the coverage parameter.
+
+    Coverage formats (same as ART):
```
ckirkman-IDM, line 570: As noted elsewhere, please include default behaviors in here.

stisim/interventions/hiv_interventions.py
```diff
@@ -13,6 +15,210 @@
...
+    Handle deprecated coverage_data and future_coverage kwargs.
+    Returns the normalized coverage input and any remaining future_coverage.
+    """
+    future_coverage = kwargs.pop('future_coverage', None)
+
+    if coverage_data is not None and coverage is not None:
+        raise ValueError('Cannot specify both coverage and coverage_data. Use coverage only.')
+
+    if coverage_data is not None:
+        warnings.warn('coverage_data is deprecated; use coverage instead', FutureWarning, stacklevel=3)
```
ckirkman-IDM, line 235: stacklevel=3! I've always used stacklevel=2 (but it is valid here, fun example :)

stisim/interventions/hiv_interventions.py
```diff
@@ -13,6 +15,210 @@
...
+                vals  = subset[val_col].values.astype(float)
+                mask = ~np.isnan(vals)
+                if mask.any():
+                    vals = np.interp(years, years[mask], vals[mask])
+                coverage[(ab, sex)] = sc.smoothinterp(yearvec, years, vals, smoothness=0)
+
+    return coverage, fmt, age_bins, sex_keys
+
+
+def _handle_deprecated_coverage(coverage, coverage_data, kwargs):
```
ckirkman-IDM, line 224: Is there a reason why coverage related parameters are not in a superclass somewhere? It seems to me that a singular interface for coverage information specification/processing would be great (rather than close to the leaf classes, as here) (another potential issue, not for this PR/issue.)

stisim/interventions/hiv_interventions.py
```diff
@@ -31,7 +237,44 @@ def administer(self, sim, uids):
...
+
+    The testing → diagnosis → ART pipeline works as follows:
+
+        1. HIVTest tests eligible agents each timestep (annual rate, scaled by dt)
+        2. Positive results set hiv.diagnosed=True and hiv.ti_diagnosed
+        3. ART checks for newly diagnosed agents (ti_diagnosed == current ti)
+        4. Newly diagnosed agents initiate ART with probability art_initiation
+        5. If coverage data is provided, ART corrects to match targets
+
+    Args:
```
ckirkman-IDM, line 276: Defaults, as always :)

stisim/interventions/hiv_interventions.py
```diff
@@ -50,56 +293,151 @@ def step(self, uids=None):
...
-        if data is not None:
-            if (len(data.columns) > 1) or (data.columns[0] not in ['n_art', 'p_art']):
-                errormsg = 'Expecting a dataframe with a single column labeled n_art or p_art'
-                raise ValueError(errormsg)
-            colname = data.columns[0]
-            self.coverage_format = colname
-            self.coverage = sc.smoothinterp(self.t.yearvec, data.index.values, data[colname].values)
+
+        # Warn if no HIV testing intervention found
+        has_hiv_test = any(isinstance(m, HIVTest) for m in sim.interventions())
```
ckirkman-IDM: Should the check be for an HIVTest BEFORE this intervention, in sequence?

stisim/interventions/hiv_interventions.py
```diff
@@ -50,56 +293,151 @@ def step(self, uids=None):
...
+
+        Returns None if no coverage data is provided (meaning: no capacity
+        constraint, just treat everyone who initiates). Returns an int otherwise.
+        """
+        # Legacy future_coverage mode
+        if self._future_coverage is not None and self.t.now('year') >= self._future_coverage['year']:
+            p_cov = self._future_coverage['prop']
+            return int(p_cov * len(eligible_uids))
+
+        if self.coverage is None:
```
ckirkman-IDM, line 421: Same None/0 comment in another class.

tests/test_hiv_interventions.py
```diff
@@ -0,0 +1,441 @@
...
+    results = dict()
+    for label, art in specs.items():
+        sim = hivsim.demo('simple', run=False, plot=False, n_agents=n_agents)
+        sim.pars.interventions = [sti.HIVTest(name='hiv_test', test_prob_data=0.3), art]
+        sim.run()
+        n_art = sim.results.hiv.n_on_art[-1]
+        results[label] = n_art
+
+    for label, n_art in results.items():
+        assert n_art > 0, f'Expected people on ART with {label} format, got {n_art}'
```
ckirkman-IDM, line 49: Hm, is it possible to test an exact count of agents on ART? Or should that be a separate test/test suite verifying targeting behavior?

tests/test_hiv_interventions.py
```diff
@@ -0,0 +1,441 @@
...
+
+@sc.timer()
+def test_vmmc_specs(do_plot=do_plot):
+    """ Check that scalar and dict VMMC coverage produce circumcisions """
+    sc.heading('Testing VMMC coverage formats...')
+
+    specs = dict(
+        scalar = sti.VMMC(coverage=0.3),
+        dict_p = sti.VMMC(coverage={'year': [2000, 2020], 'value': [0, 0.3]}),
+        df_p   = sti.VMMC(coverage=pd.DataFrame(index=np.arange(2000, 2021), data={'p_vmmc': np.linspace(0, 0.3, 21)})),
```
ckirkman-IDM, line 103: I need to level up my one-liner dataframe making skills!

tests/test_hiv_interventions.py
```diff
@@ -0,0 +1,441 @@
...
+        sim_without = sti.Sim(
+            diseases=sti.HIV(beta_m2f=0.05, init_prev=0.05),
+            networks=sti.StructuredSexual(),
+            demographics=[ss.Pregnancy(fertility_rate=10), ss.Deaths(death_rate=10)],
+            interventions=[sti.ART(coverage=0.9)],
+            n_agents=n_agents, dur=20, verbose=-1,
+        )
+        sim_without.run()
+        warn_texts = [str(x.message) for x in w]
+        has_testing_warn = any('testing' in t.lower() for t in warn_texts)
```
ckirkman-IDM, line 179: nit: is this specific enough?

tests/test_hiv_interventions.py
```diff
@@ -0,0 +1,441 @@
...
+    )
+    sim.run()
+    assert sim.results.hiv.n_on_art[-1] > 0, 'Expected people on ART without pregnancy module'
+
+    return sim
+
+
+# %% Coverage matching tests
+@sc.timer()
+def test_art_coverage_matching(do_plot=do_plot):
```
ckirkman-IDM, line 219: Ah, actual art coverage testing :)

tests/test_hiv_interventions.py
```diff
@@ -0,0 +1,441 @@
...
+def test_art_stratified_coverage_matching(do_plot=do_plot):
+    """ Check that age-stratified ART coverage approximately matches supplied targets """
+    sc.heading('Testing stratified ART coverage matching...')
+
+    # Build stratified coverage data: higher coverage for older age groups and women
+    age_bins = [15, 25, 35, 45]
+    targets = {}
+    rows = []
+    for year in [2005, 2015]:
+        for gender in [0, 1]:
```
ckirkman-IDM, line 264: Is integer gender documented somewhere? Might be nice to unify gender parsing (int/str/other) somewhere someday.

tests/test_hiv_interventions.py
```diff
@@ -0,0 +1,441 @@
...
+            ax.set_ylabel('ART coverage')
+            ax.set_title(f'{"Women" if sex == "f" else "Men"}')
+            ax.legend()
+            ax.set_ylim(0, 1)
+
+    return sim
+
+
+# %% Scientific validation skeletons
+@sc.timer()
```
ckirkman-IDM, line 315: Do you see this test as sufficient to resolve:
#241 ?

Oh, I have a update in #241! Ok, separate tests to approach the same question, like a combinatorial argument! Excellent.

tests/test_hiv_interventions.py
```diff
@@ -0,0 +1,441 @@
...
+        ax.set_title('ART mortality reduction')
+
+    # TODO: uncomment once parameterized and validated with sufficient population
+    # assert deaths_art < deaths_no_art, f'Expected fewer deaths with ART ({deaths_art}) than without ({deaths_no_art})'
+
+    return sim_no_art, sim_art
+
+
+@sc.timer()
+def test_art_parameter_sensitivity(do_plot=do_plot):
```
ckirkman-IDM, line 351: Go ahead and claim #236 if this is intended to address it! (and/or reword it to include these parameter variations, too)

tests/test_hiv_interventions.py
```diff
@@ -0,0 +1,441 @@
...
+    #     v_lo = results[(par, lo)]
+    #     v_hi = results[(par, hi)]
+    #     assert v_lo <= v_hi, f'Expected higher {par} to increase ART uptake, but {par}={lo} gave {v_lo} vs {par}={hi} gave {v_hi}'
+
+    return results
+
+
+@sc.timer()
+def test_art_duration(do_plot=do_plot):
+    """ Check that agents remain on ART for expected durations """
```
ckirkman-IDM, line 389: High level feedback on unfinished test :)

If this is done per-agent, is the information available to do:
time_art_drop_out - time_art_start = dur_on_art
... with all mortality turned off (including HIV, if possible).

... then do a check to ensure proper duration for all agents. I know, using an analyzer .... :D

ckirkman-IDM (changes requested):
Ok, lots of comments.

The "required" ones for merging are basically the documentation of default behaviors. 

Anything else noted as "another issue", "maybe we could do, ...", etc. are future thoughts. Feel free to make any issue(s) for discussion out of those if you feel they are worthwhile for discussion later on.

If you have any follow-on questions or want to discuss any of this, let me know!

stisim/interventions/base_interventions.py
robynstuart: Done

stisim/interventions/base_interventions.py
robynstuart: Added

stisim/interventions/base_interventions.py
robynstuart: Great catch. The old linear approximation (p * dt) was wrong — test_prob_data=0.24 with monthly dt gave 21.5%/year instead of 24%. Now uses ss.probperyear(test_prob).to_prob(sim.dt) for correct exponential conversion. Updated all docs to say "annual probability" not "annual rate".

stisim/interventions/base_interventions.py
robynstuart: Added

stisim/interventions/base_interventions.py
robynstuart: Added

stisim/interventions/base_interventions.py
robynstuart: Documented with example. Default: all agents (HIVTest overrides to undiagnosed only).

stisim/interventions/base_interventions.py
robynstuart: Documented. HIVTest auto-creates a perfect HIVDx; STITest requires one.

docs/user_guide/interventions/interventions.md
robynstuart: Documented, targets all males by default, eligibility parameter available.

docs/user_guide/interventions/interventions.md
robynstuart: Added to key points and ART docstring. `ART.init_pre` now warns if HIVTest comes after ART in the list.

docs/user_guide/interventions/interventions.md
robynstuart: Clarified in the comment, agents are added/removed immediately (not gradually) each timestep.

docs/user_guide/interventions/interventions.md
robynstuart: All defaults are now in both the user guide and the class docstrings.

docs/user_guide/interventions/interventions.md
robynstuart: All defaults now documented in docstrings and user guide.

docs/user_guide/interventions/interventions.md
robynstuart: Targets HIV-negative FSWs by default, coverage/years defaults shown, eligibility documented.

stisim/analyzers.py
robynstuart: Added documentation that these are [lo, hi) half-open intervals, also in the `_parse_age_bin` docstring

docs/user_guide/interventions/interventions.md
ckirkman-IDM: Not to be pedantic, but thinking about this more ... in a scenario where > 1 HIVTest exists in the interventions (for example, one for fsw and one for pregnant mothers-to-be, there may be > 1 ART intervention keyed to those populations, too. (maybe pregnant women and fsw have different uptake rates after being diagnosed). How would we keep track of ordering in this case?

I think the answer is "we can't" in a very general sense. So we will need to rely on training our users to think like the model (in addition to any warnings we CAN provide).

stisim/analyzers.py
robynstuart: Yeah... I was looking through https://github.com/starsimhub/starsim/issues/1131, and we did at some point consider having a `Par` class that would store labels and other data for each par. But we haven't taken that path, so parameters are typically just described in docstrings. Makes it all the more important to have good docstrings/comments!

stisim/analyzers.py
robynstuart: Noted! Will try to guard against it with the docs

stisim/interventions/hiv_interventions.py
robynstuart: Good point, `_parse_age_bin` strips all brackets and treats everything as [lo, hi). Made an issue: https://github.com/starsimhub/stisim/issues/326

stisim/interventions/hiv_interventions.py
robynstuart: https://github.com/starsimhub/stisim/issues/327

stisim/interventions/hiv_interventions.py
robynstuart: https://github.com/starsimhub/stisim/issues/327

stisim/interventions/hiv_interventions.py
robynstuart: The distinction matters for ART (None = no constraint, 0 = remove everyone from ART). For VMMC it's less critical since VMMC is one-way (can't un-circumcise), but keeping it consistent with ART for now.

stisim/interventions/hiv_interventions.py
robynstuart: https://github.com/starsimhub/stisim/issues/328

stisim/interventions/hiv_interventions.py
robynstuart: Done

stisim/interventions/hiv_interventions.py
robynstuart: https://github.com/starsimhub/stisim/issues/328

stisim/interventions/hiv_interventions.py
robynstuart: Added!

stisim/interventions/hiv_interventions.py
robynstuart: Done

stisim/interventions/hiv_interventions.py
robynstuart: Addressed in comment above, but would also maybe be fixed by https://github.com/starsimhub/stisim/issues/328...

tests/test_hiv_interventions.py
robynstuart: It's included below!

tests/test_hiv_interventions.py
robynstuart: Think it's ok for here

tests/test_hiv_interventions.py
robynstuart: good point. Added more functions here: https://github.com/starsimhub/stisim/blob/intvs-docs/stisim/interventions/hiv_interventions.py#L111-L175

stisim/interventions/hiv_interventions.py
robynstuart: Very true. The Prep intervention here is super skeletal. I just added a group of PrEP-related issues here: https://github.com/starsimhub/stisim/issues/329

stisim/interventions/base_interventions.py
ckirkman-IDM: Can you point me to this change, I don't see it :/ . Or is it in a newly created issue? Update: Found it!

stisim/interventions/hiv_interventions.py
ckirkman-IDM: I think consistency wins, then. I think ART and VMMC may converge in their usage in this regard when "remove everyone from ART" (ART tracking) becomes an option; it would be good to think about converging meanings of None/0/etc as much as possible for intuition building. (I admit, I also like constant/enum values for syntactic sugar ... e.g. 

```
UNCONSTRAINED = None 
...
if n_to_circ is not UNCONSTRAINED or n_to_circ > 0:
  ...
```

Anyway, it is fine for now.

stisim/diseases/sti.py
```diff
@@ -511,6 +511,10 @@ def set_exposure(self, uids):
         return
 
     def set_pars(self, par, uids):
+        # Handle scalar calibration values by converting to [scalar, scalar]
+        if sc.isnumber(par):
+            par = [par, par]
+
```
ckirkman-IDM, line 517: Related to this PR? (I'm not a purist, it's ok if needed, just making sure something didn't slip in)

stisim/interventions/base_interventions.py
```diff
@@ -178,7 +194,8 @@ def make_test_prob_fn(self, sim, uids):
 
         # Scale and validate
         test_prob *= self.pars.rel_test
-        if self.pars.dt_scale: test_prob *= self.t.dt_year
+        if self.pars.dt_scale:
+            test_prob = ss.probperyear(test_prob).to_prob(sim.dt)
```
ckirkman-IDM, line 198: Ah, found it in "new since last view" view :D

ckirkman-IDM (commented):
One last question (all else is fine, I think):

This comment is it, in sti.py:
Related to this PR? (I'm not a purist, it's ok if needed, just making sure something didn't slip in)

stisim/interventions/base_interventions.py
robynstuart: It's here: https://github.com/starsimhub/stisim/blob/intvs-docs/stisim/interventions/base_interventions.py#L198

PR #338, opened 2026-03-12
Ensuring no HIV infections exist in the absence of seeding/events driving them (Fix #271)

### LINKED ISSUE #271: Test zero infections scenario
Sub-issue of #201.

ckirkman-IDM:


tests/hiv_natural_history_analyzers.py
```diff
@@ -35,24 +36,39 @@ def update_results(self):
...
     """
-    Records an agent-uid-keyed dict of timeseries of CD4 count. Results obtainable by analyzer key 'hiv.ts_cd4'
+    Records an agent-uid-keyed dict of timeseries of CD4 count for specified subpopulation. Results obtainable by
+    analyzer key 'hiv.ts_cd4' .
     """
+    result_name = 'hiv.ts_cd4'
+
+    INFECTED = 'infected'
+    ONART = 'onart'
+    SUBPOPS = {
```
robynstuart: this SUBPOPS dict-of-lambdas pattern is hard to read at a glance. Since there are only two cases, this could be much simpler:

```
class CD4ByUIDTracker(ss.Analyzer):
    """Records per-agent CD4 timeseries for infected or on-ART agents."""
    
    result_name = 'hiv.ts_cd4'

    def __init__(self, track_art=False, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.track_art = track_art

    # ...

    def update_results(self):
        hiv = self.sim.diseases.hiv
        uids = hiv.on_art.uids if self.track_art else hiv.infected.uids
        for uid in uids:
            if uid not in self.results[self.result_name]:
                self.results[self.result_name][uid] = []
            self.results[self.result_name][uid].append(hiv.cd4[uid])
```

tests/hiv_natural_history_analyzers.py
```diff
@@ -35,24 +36,39 @@ def update_results(self):
...
-        self.results['hiv.ts_cd4'] = {}
+        self.results[self.result_name] = {}
 
     def update_results(self):
         hiv = self.sim.diseases.hiv
-        infected_uids = hiv.infected.uids
-        for uid in infected_uids:
-            if uid not in self.results['hiv.ts_cd4']:
-                self.results['hiv.ts_cd4'][uid] = []
+        uids = self.SUBPOPS[self.subpop]['filter'](hiv=hiv)
```
robynstuart: ditto, hard to read: could be `uids = hiv.on_art.uids if self.track_art else hiv.infected.uids`. Getting picky on this because I think we might want to make this a reusable analyzer: https://github.com/starsimhub/stisim/issues/340

tests/hiv_natural_history_analyzers.py
```diff
@@ -141,6 +157,29 @@ def update_results(self):
         self.results[self.result_name][self.ti] = transmissions
 
 
+class PrevalenceTracker(ss.Analyzer):
```
robynstuart: is this needed? The HIV module already computes prevalence every timestep as a built-in result. So test_no_hiv_with_no_outbreaks could just use:

```
sim.run()
prevalence = sim.results.hiv.prevalence[:]
assert np.all(prevalence == 0), "Expected zero prevalence without seeding"
```
Could you check whether the built-in result works for your test? If there's a reason it doesn't, happy to discuss.

tests/test_hiv_natural_history_verification.py
```diff
@@ -231,6 +234,91 @@ def test_mtc_transmission_occurs():
...
+def test_cd4_rises_on_ART():
+    sc.heading("Ensuring that agent CD4 levels rise when on ART (monotonically in a short test)")
+
+    # To keep test small, infect everyone with HIV and put everyone on ART immediately
+    analyzer = CD4ByUIDTracker(subpop=CD4ByUIDTracker.ONART)
+    test_intervention = HIVTest(test_prob_data=1.0, dt_scale=False)  # everyone tests, first timestep
+    initial_art_intervention = ART(art_initiation=1.0)  # everyone diagnosed starts ART.
+
+    # agents are on art for the full sim length
+    #   ... and more (due to bug: https://github.com/starsimhub/stisim/issues/336)
```
robynstuart: workarounds can be simplified now this bug has been fixed, so `hiv = sti.HIV(..., dur_on_art=ss.constant(v=ss.years(5)))` should work.

robynstuart (changes requested):
Thanks @ckirkman-IDM! A few suggestions before merging. No rish on these, can bring them into v151

tests/hiv_natural_history_analyzers.py
ckirkman-IDM: Never know, it might be more ... I have updated the pattern, though I do like having the logic at the top, away from having to read code further down.

tests/hiv_natural_history_analyzers.py
ckirkman-IDM: Thanks, found the default reporting after your comment. There is no need to use a prevalence tracker analyzer here, as the built-in n_infections counting will work.

tests/test_hiv_natural_history_verification.py
ckirkman-IDM: done.

PR #346, opened 2026-03-17
HIV: Adding tests to ensure that dropping ART results in decreasing cd4 counts and increasing transmissability (Fix #244)

### LINKED ISSUE #244: Test rel_trans rebound after treatment interruption
Sub-issue of #195.

ckirkman-IDM:


tests/hiv_natural_history_analyzers.py
```diff
@@ -78,6 +78,51 @@ def update_results(self):
             self.results[result_name][uid].append(cd4)
 
 
+class RelativeInfectivityByUIDTracker(ss.Analyzer):
```
robynstuart, line 81: Not needed for this PR, but given how common the pattern is now of tracking results by UID, I wonder whether we should make this a more general class that specific analyzers can inherit from

tests/hiv_natural_history_analyzers.py
```diff
@@ -78,6 +78,51 @@ def update_results(self):
...
+    """
+    Records an agent-uid-keyed dict of timeseries of CD4 count for specified subpopulation. Results obtainable by
+    analyzer key 'hiv.ts_cd4' .
+    """
+    result_name = 'hiv.ts_rel_trans'
+
+    INFECTED = 'infected'
+    ONART = 'onart'
```
robynstuart: wouldn't it be simpler to match the attributes that are recorded in the HIV module, so on_art rather than onart?

robynstuart (commented):
Thanks @ckirkman-IDM! Looking at the new RelativeInfectivityByUIDTracker alongside the existing CD4ByUIDTracker, I notice they're almost identical, with the only differences are (1) which subpopulation to track and (2) which value to record per agent. The same pattern shows up in RelativeInfectivityTracker too.

What would you think about consolidating these into a single generic track_by analyzer? Something like:

```
tracker = sti.track_by(
    subpop=lambda sim: sim.diseases.hiv.infected & sim.diseases.hiv.on_art,
    value=lambda sim, uids: sim.diseases.hiv.rel_trans[uids],
)
```
The subpop argument would follow the same pattern we use for eligibility in the interventions: either a function that takes sim and returns UIDs/BoolArr, or a static set of UIDs. The value argument would be a function that extracts the quantity to record for those agents.

That way the three tests could reuse one analyzer class:

```
cd4_tracker = sti.track_by(
    subpop=lambda sim: sim.diseases.hiv.infected,
    value=lambda sim, uids: sim.diseases.hiv.cd4[uids],
)

rt_tracker = sti.track_by(
    subpop=lambda sim: sim.diseases.hiv.on_art,
    value=lambda sim, uids: sim.diseases.hiv.rel_trans[uids],
)
```

This could live in stisim/analyzers.py and would replace the need for test-specific analyzer classes. Happy to help build it if you think it's a good direction!

Another important thing to note: biologically, someone who drops off ART and has declining CD4 would eventually become more infectious again. But that's not happening here, we should make an issue about rel_trans.

tests/hiv_natural_history_analyzers.py
ckirkman-IDM: DEFINITELY. I played around with exactly the idea last Friday but didn't settle on a final superclass yet.

tests/hiv_natural_history_analyzers.py
ckirkman-IDM: I can swap 'onart' to 'on_art'. I'm used to OnART from dtk-tools land :D. I do like consistencty.

ckirkman-IDM: Done.

ckirkman-IDM:
> This could live in stisim/analyzers.py and would replace the need for test-specific analyzer classes. Happy to help build it if you think it's a good direction!

We should do this, but in another issue (and refactor the tests along with).

 > Another important thing to note: biologically, someone who drops off ART and has declining CD4 would eventually become more infectious again. But that's not happening here, we should make an issue about rel_trans.

There are two tests in this PR:
- cd4 falls after ART dropout
- rel_trans rises after ART dropout

Unless there is a timing issue we want to check (rel_trans rises in sync or after cd4 fall, or after some dt), this should cover your note. Also unless I misread your comment, which is possible, too.

PR #347, opened 2026-03-17
Adding test to verify increased testing rate leads to more/faster diagnoses of HIV (Fix #253)

### LINKED ISSUE #253: Test that increased testing leads to earlier diagnosis
Sub-issue of #197.

ckirkman-IDM:


robynstuart (commented):
Looks good, only comment is that we could run the 2 sims in parallel

ckirkman-IDM:
Now in parallel.

PR #348, opened 2026-03-17
Adding test to check exponentiality of early-phase HIV epidemics (linearity of log(n_infections)) (Fix #260)

### LINKED ISSUE #260: Test exponential growth in early epidemic
Sub-issue of #198.

ckirkman-IDM:
Made an assumption of 0.9 correlation / 0.81 variance explained threshold. Let me know if my "early" epidemic assumptions should be updated.

### DIFF
--- a/tests/devtests/devtest_hiv.py
+++ b/tests/devtests/devtest_hiv.py
@@ -5,6 +5,17 @@ import stisim as sti
 import hivsim
 import matplotlib.pyplot as plt
 
+from statistics import log
+from pathlib import Path
+import sys
+
+tests_directory = Path(__file__).resolve().parent.parent  # up to stisim/tests/ dir
+sys.path.append(str(tests_directory))
+
+from testlib import build_testing_sim
+
+verbose = False
+
 
 class TrackValues(ss.Analyzer):
     # Track outputs for viral load and CD4 counts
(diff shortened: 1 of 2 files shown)


robynstuart (commented):
Looks good! Do you feel confident that it's robust enough with threshold=0.9? Could remove the assertion and just include this test as a gut check if you prefer

ckirkman-IDM:
> Looks good! Do you feel confident that it's robust enough with threshold=0.9? Could remove the assertion and just include this test as a gut check if you prefer

Good question. I meant to run a sweep of random seeds to verify, but it is definitely not consistent at the current settings. I played with them to get a high correlation and still be reasonably "early" phase. I just tried 100 seeds and only a quarter pass the threshold, some failing miserably. There is definitely a seeding/network issue. This may be a good candidate for either running with multiple random seeds and asserting some fraction or better look good, OR moving into a suite of "gut check" like statistics/graphs test suite with no asserts.

I'll play with it a bit more and see if I can make it reasonably robust/short. I'll report back here either way.

Update: I've gotten it to run successfully for 1000 random seeds, but have pared it to 10 seeds + moved it to devtest_hiv.py with no asserts (just prints) as discussed. Will merge momentarily when green light appears.

PR #351, opened 2026-03-19
Updates for hiv_kenya

cliffckerr:
Updates to parameter handling that were encountered when working with hiv_kenya. NB: I haven't incremented the version or any of that admin!

### DIFF
--- a/stisim/demographics.py
+++ b/stisim/demographics.py
@@ -21,7 +21,7 @@ class Migration(ss.Demographics):
         super().__init__()
         self.define_pars(
             # Migration parameters
-            dt = ss.years(1),  # Unit of time for migration data
+            # dt = ss.years(1),  # Unit of time for migration data # TODO: not implemented
             migration_propensity = ss.normal(loc=1, scale=0.1),  # Propensity to emigrate
             slot_scale = 5, # Random slots will be assigned to newborn agents between min=n_agents and max=slot_scale*n_agents
             min_slots  = 100, # Minimum number of slots, useful if the population size is very small
@@ -124,7 +124,7 @@ class Migration(ss.Demographics):
             # Choose people to twin
             choices = np.argsort(-weights)[:new_migrants]
             twin_uids = uids.uids[choices]
-            new_people = self.make_immigrants(twin_uids)
+            self.make_immigrants(twin_uids)
 
         # new_migrants is negative -> remove people from the population
         elif new_migrants < 0:
--- a/stisim/parameters.py
+++ b/stisim/parameters.py
@@ -1,7 +1,6 @@
 """
 Set parameters
 """
-import numpy as np
 import sciris as sc
 import starsim as ss
 import stisim as sti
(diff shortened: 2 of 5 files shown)


stisim/demographics.py
```diff
@@ -21,7 +21,7 @@ def __init__(self, pars=None, migration_data=None, **kwargs):
         super().__init__()
         self.define_pars(
             # Migration parameters
-            dt = ss.years(1),  # Unit of time for migration data
+            # dt = ss.years(1),  # Unit of time for migration data # TODO: not implemented
```
robynstuart, line 24: what's not implemented exactly?

stisim/parameters.py
```diff
@@ -20,14 +19,8 @@ def __init__(self, **kwargs):
         # Initialize the parent class
         super().__init__()
 
-        # General parameters
-        self.label   = ''  # The label of the simulation
-        self.verbose = ss.options.verbose  # Whether or not to display information during the run -- options are 0 (silent), 0.1 (some; default), 1 (default), 2 (everything)
-
         # Population parameters
         self.n_agents  = 1e3   # Number of agents
-        self.total_pop = None  # If defined, used for calculating the scale factor
```
robynstuart, line 29: have these been removed because they're all identical to the starsim defaults?

stisim/sim.py
```diff
@@ -111,38 +130,56 @@ def separate_pars(self, pars=None, sim_pars=None, sti_pars=None, nw_pars=None, d
         # Merge in pars and kwargs
         all_pars = sc.mergedicts(pars, sim_pars, sti_pars, nw_pars, dem_pars, sim_kwargs, kwargs)
         all_pars = self.remap_pars(all_pars)  # Remap any parameter names
+        vp, duplicates = self.get_valid_pars()
+
+        # Check if any parameter in all_pars is in duplicates
+        ambiguous = set(all_pars.keys()) & set(duplicates)
+        if ambiguous: # Warn if and only if this parameter is specified
+            warnmsg = f'Ambiguous parameter(s) found: {ambiguous}'
```
robynstuart, line 138: Do we want to make an issue for this?

robynstuart (commented):
Thank you @cliffckerr! Minor comments below. Can you rebase against rc1.5.1 and merge when you're ready?

stisim/parameters.py
cliffckerr: yes exactly

stisim/sim.py
cliffckerr: I thought I fixed all the ambiguous parameters, but if not then yes!

stisim/demographics.py
cliffckerr: `self.pars.dt` is not used to scale the migration data, that I could see? But in any case doesn't this collide with the module timestep, so shouldn't it be something like `data_dt` in case the data time unit and the module update frequency are different?

PR #384, opened 2026-04-08
Replace _parse_age_bin with ss.parse_age_range (#326)

robynstuart:
## NOTE
GHA will fail because this branch has upstream dependencies on Starsim changes. 

## Summary
- Deletes `_parse_age_bin` from `hiv_interventions.py` and replaces all usage with `ss.apply_age_range` (bracket-aware) and `ss.parse_age_range`
- Fixes lexicographic sort bug in `_parse_stratified_df` — age bins now sorted numerically
- Bumps starsim dependency to `>=3.3.3` (requires starsimhub/starsim `age-ranges` branch)

Closes #326

Merge starsim `age-ranges` branch first (starsimhub/starsim `age-ranges`)

PR #387, opened 2026-04-08
Extract shared coverage-targeting logic (#328)

### LINKED ISSUE #328: Shared coverage-targeting logic
ART, VMMC, and (partially) PrEP all share the same pattern for coverage-based targeting:
- Parse coverage data in init_pre via parse_coverage()
- Store self.coverage, self.coverage_format, self.age_bins, self.sex_keys
- Compute a target number each timestep from coverage data (aggregate or stratified)
- do something to loop over age_bins/sex_keys, same coverage lookup, same stratum filtering with _parse_age_bin

This logic is currently duplicated across classes. 

Related: #326 (consistent interpolation)

robynstuart:
## Summary
- New `stisim/interventions/utils.py` with coverage parsing and targeting helpers extracted from `hiv_interventions.py`
- `age_sex_mask`, `coverage_to_number`, `compute_coverage_target` as composable shared functions
- ART and VMMC coverage methods become one-line delegates
- Age-only stratification support (no Gender column required)
- Remove duplicate `count()` from `base_interventions.py` and `gonorrhea_interventions.py`

Depends on #384 (`age-ranges` branch)
Closes #328

### DIFF
--- a/stisim/interventions/base_interventions.py
+++ b/stisim/interventions/base_interventions.py
@@ -7,10 +7,7 @@ import numpy as np
 import sciris as sc
 import pandas as pd
 from scipy.interpolate import interp1d
-
-
-# %% Helper functions
-def count(arr): return np.count_nonzero(arr)
+from stisim.utils import count
 
 
 # %% Base classes
--- a/stisim/interventions/utils.py
+++ b/stisim/interventions/utils.py
@@ -1,0 +1,345 @@
+"""
+Shared utilities for STIsim interventions.
+
+Coverage parsing, stratified targeting, and related helpers used across
+intervention classes (ART, VMMC, etc.).
+"""
+
+import warnings
+import starsim as ss
+import numpy as np
+import pandas as pd
+import sciris as sc
+
+
+# %% Coverage parsing
+
+def parse_coverage(data, valid_names=None, yearvec=None):
+    """
+    Parse coverage data into a per-timestep array.
+
+    Returns (coverage_array, coverage_format, age_bins, sex_keys) where:
+        - coverage_array: 1D array (len=yearvec) for aggregate, or dict of arrays for stratified
+        - coverage_format: 'n' (absolute numbers) or 'p' (proportion)
+        - age_bins: list of age bin strings if stratified, else None
+        - sex_keys: list of sex values if stratified, else None
+
+    Examples::
+
+        # Constant proportion
+        parse_coverage(0.9, yearvec=yearvec)
+
+        # Time-varying dict (format auto-detected from values)
+        parse_coverage({'year': [2000, 2020], 'value': [0, 0.9]}, yearvec=yearvec)
+
+        # Time-varying dict with explicit format
+        parse_coverage({'year': [2000, 2020], 'value': [1000, 50000], 'format': 'n'}, yearvec=yearvec)
+
+        # Single-column DataFrame
+        parse_coverage(pd.DataFrame(index=years, data={'n_art': vals}), valid_names=['n_art'], yearvec=yearvec)
+
+        # Age/sex stratified DataFrame (Gender column is optional)
+        parse_coverage(strat_df, valid_names=['p_art'], yearvec=yearvec)
+
+    Args:
+        data:        coverage input in any of the formats above, or None
+        valid_names: list of valid column names, e.g. ['n_art', 'p_art']
+        yearvec:     simulation year vector for interpolation
+    """
+    if valid_names is None:
+        valid_names = []
+
+    # None → no coverage
+    if data is None:
+        return None, None, None, None
+
+    # Scalar → constant proportion
+    if sc.isnumber(data):
+        arr = np.full(len(yearvec), float(data))
+        return arr, 'p', None, None
+
+    # Dict with year + value keys → interpolate
+    if isinstance(data, dict) and 'year' in data and 'value' in data:
+        years  = np.array(data['year'], dtype=float)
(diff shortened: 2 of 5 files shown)


stisim/interventions/base_interventions.py
```diff
@@ -7,10 +7,7 @@
 import sciris as sc
 import pandas as pd
 from scipy.interpolate import interp1d
-
-
-# %% Helper functions
-def count(arr): return np.count_nonzero(arr)
+from stisim.utils import count
```
cliffckerr, line 10: For `ss.Arr`, you can just use `.count()` directly, but I see not all of these are Starsim arrays. Boo to NumPy for making `count_nonzero()` so clunky! (And also boo to them for making `arr.sum()` so much slower for Boolean arrays when it could just call `np.count_nonzero(arr)` underneath.)

stisim/interventions/utils.py
```diff
@@ -0,0 +1,345 @@
...
+import numpy as np
+import pandas as pd
+import sciris as sc
+
+
+# %% Coverage parsing
+
+def parse_coverage(data, valid_names=None, yearvec=None):
+    """
+    Parse coverage data into a per-timestep array.
```
cliffckerr, line 19: I know you have _examples_ of what `data` is, but a longer docstring saying what the principles behind it are would be good too.

stisim/interventions/utils.py
```diff
@@ -0,0 +1,345 @@
...
+    rename = {}
+    col_lower = {c.lower(): c for c in data.columns}
+
+    # Year
+    for alias in ['year']:
+        if alias in col_lower:
+            rename[col_lower[alias]] = 'Year'
+            break
+    else:
+        raise ValueError(f'Stratified coverage DataFrame must have a "Year" column. Found: {list(data.columns)}')
```
cliffckerr: I prefer to separate out `errormsg` onto a separate line for any nontrivial error messages (especially f-string ones) just because otherwise Python duplicates the whole thing for reasons that I don't understand:
```py
>>> raise ValueError("I really don't understand why Python duplicates the whole error message when you raise an exception")

Traceback (most recent call last):
  File "<python-input-0>", line 1, in <module>
    raise ValueError("I really don't understand why Python duplicates the whole error message when you raise an exception")
ValueError: I really don't understand why Python duplicates the whole error message when you raise an exception
```

stisim/interventions/utils.py
```diff
@@ -0,0 +1,345 @@
...
+
+    # Sex/Gender (optional — age-only stratification is valid, e.g. for VMMC)
+    has_sex = False
+    for alias in ['gender', 'sex']:
+        if alias in col_lower:
+            rename[col_lower[alias]] = 'Gender'
+            has_sex = True
```
cliffckerr: This seems a little confusing to me ... I guess I could see some places where we would want to use sex (e.g., transmission risk), and other places where we would want to use gender (e.g., SES factors), so I'm agnostic about whether "sex" or "gender" makes more sense as the "canonical" term (though I would kind of lean towards the former). However, I do have a problem with making "Gender" the canonical term and then calling the variable `has_sex`!

stisim/interventions/utils.py
```diff
@@ -0,0 +1,345 @@
...
+    # Optional columns
+    for alias, canonical in [('count', 'Count'), ('lb', 'lb'), ('ub', 'ub')]:
+        if alias in col_lower:
+            rename[col_lower[alias]] = canonical
+
+    df = data.rename(columns=rename)
+
+    # Normalize gender values to integers: 0=female, 1=male
+    if has_sex:
+        df['Gender'] = df['Gender'].map(_normalize_sex)
```
cliffckerr: I can also see some questions arising from talking about gender and then mapping it onto a literal binary :) (I appreciate that you do say int, not bool, though `_normalize_sex()` only currently handles `m` or `f`.)

stisim/interventions/utils.py
```diff
@@ -0,0 +1,345 @@
...
+                vals  = subset[val_col].values.astype(float)
+                mask = ~np.isnan(vals)
+                if mask.any():
+                    vals = np.interp(years, years[mask], vals[mask])
+                coverage[key] = sc.smoothinterp(yearvec, years, vals, smoothness=0)
+
+    return coverage, fmt, age_bins, sex_keys
+
+
+def _handle_deprecated_coverage(coverage, coverage_data, kwargs):
```
cliffckerr, line 258: Do we have few enough users that we can just remove this?

stisim/interventions/utils.py
```diff
@@ -0,0 +1,345 @@
...
+        BoolArr: True for agents in the stratum
+    """
+    mask = ss.apply_age_range(age_bin, people.age)
+    if sex is not None:
+        sex_mask = people.female if sex == 0 else people.male
+        mask = mask & sex_mask
+    return mask
+
+
+def coverage_to_number(cov_val, coverage_format, pop_scale=None, n_eligible=None):
```
cliffckerr: Are these functions for the user? If not, consider starting with `_`.

cliffckerr (commented):
Looks good, some questions about sex vs gender

stisim/interventions/utils.py
ckirkman-IDM: Python isn't distinguishing when code you wrote vs someone else wrote throws an exception. It just shows you the line that threw it, which may not be descriptive at all, and the generated exception & message (which STILL may not be very descriptive! :D).

PR #388, opened 2026-04-09
Consistent interpolation and PrEP refactor (#327)

### LINKED ISSUE #327: Consistent and flexible interpolation
Coverage and testing data are interpolated to the sim's yearvec in several places, but the interpolation method is inconsistent. Some use piecewise-constant (sc.smoothinterp(..., smoothness=0)), others use linear (np.interp). Users have no way to control which method is used.

Location | File | Line(s) | Method | Effect
-- | -- | -- | -- | --
Dict coverage (ART/VMMC) | hiv_interventions.py | 61 | sc.smoothinterp(smoothness=0) | Piecewise-constant
DataFrame coverage (legacy) | hiv_interventions.py | 94, 102, 104 | sc.smoothinterp(smoothness=0) | Piecewise-constant
Stratified DataFrame (NaN fill) | hiv_interventions.py | 196 | np.interp | Linear
Stratified DataFrame (to yearvec) | hiv_interventions.py | 197 | sc.smoothinterp(smoothness=0) | Piecewise-constant
PrEP coverage | hiv_interventions.py | 703 | np.interp | Linear

So ART/VMMC use piecewise-constant, PrEP uses linear, and the stratified path uses linear for NaN-filling then piecewise-constant for the final interpolation.

Tasks:
- [ ] Pick one default linear interpolation OR document why particular choices were made
- [ ] Add an optional interp parameter to parse_coverage() (e.g. interp='linear' or interp='nearest') that gets passed through from ART/VMMC constructors
- [ ] Make PrEP use the same parse_coverage() path rather than its own np.interp call
- [ ] Document the interpolation behavior in the user guide
- [ ] Fix docs - the user guide and ART/VMMC docstrings currently say "linearly interpolated" for dict/DataFrame coverage, but the code actually uses piecewise-constant (smoothness=0). Either the docs or the code needs to change.

robynstuart:
## Summary
- Add `smoothness` parameter to `parse_coverage` (and ART/VMMC constructors), with `**interp_kw` passed through to `sc.smoothinterp` for full flexibility
- Standardize all interpolation on `sc.smoothinterp` (smoothness=0 = linear, matching existing behavior)
- Refactor PrEP to use `parse_coverage` instead of inline `np.interp` — now accepts same flexible coverage inputs as ART/VMMC
- Replace `np.interp` with `sc.smoothinterp(smoothness=0)` in `STITreatment.init_pre`

Depends on #387 (`shared-coverage` branch)
Closes #327

### DIFF
--- a/stisim/interventions/base_interventions.py
+++ b/stisim/interventions/base_interventions.py
@@ -304,7 +304,7 @@ class SymptomaticTesting(STITest):
     def init_pre(self, sim):
         super().init_pre(sim)
         if self.treat_prob_data is not None:
-            self.treat_prob = np.interp(sim.yearvec, self.treat_prob_data.year.values, self.treat_prob_data.treat_prob.values)
+            self.treat_prob = sc.smoothinterp(sim.yearvec, self.treat_prob_data.year.values, self.treat_prob_data.treat_prob.values, smoothness=0)
         return
 
     def init_results(self):
--- a/stisim/interventions/hiv_interventions.py
+++ b/stisim/interventions/hiv_interventions.py
@@ -139,7 +139,7 @@ class ART(ss.Intervention):
         art = sti.ART(coverage=pd.read_csv('art_coverage.csv').set_index('year'))
     """
 
-    def __init__(self, pars=None, coverage=None, coverage_data=None, **kwargs):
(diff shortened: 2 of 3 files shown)


cliffckerr (approved):
Looks good!

PR #389, opened 2026-04-09
Mixed n/p coverage format (#383)

### LINKED ISSUE #383: Future coverage bug
This was encountered when running run_scenarios in syph_dx_zim repo, but is a general issue that needs addressing. 

At some point, we refactored ART with the goal of getting rid of the "future_coverage" input par. We did this because it was confusing - it was a dict keyed by date, and the date didn't need to be in the future. 

In stead, we were aiming to make ART more robust by making it possible to set coverage=None, in which case we skip the force-match-to-data logic.

However, it WOULD still be useful to be able to use an assumption for coverage as a % (p_art) once the data on the actual number of people on ART reaches its end year. It would also be nice to support using p_art as an input for all the years, or part of the years. We won't need to support using p_art and n_art for the SAME years, but we should be able to switch between them, with the primary use case being to switch from n_art (historical) to p_art (projected).

robynstuart:
## Summary
- Support dual-column DataFrames (both `n_art` and `p_art`) with `format_priority` parameter
- Support dict with per-entry format list: `{'format': ['n','n','p','p']}`
- `coverage_format` becomes per-timestep array when mixed; `coverage_to_number` resolves per timestep
- Remove deprecated `future_coverage` and `coverage_data` code paths
- Update ART/VMMC docstrings with full coverage format documentation
- Update Zimbabwe demo example

Depends on #388 (`consistent-interp` branch)
Closes #383

### DIFF
--- a/stisim/interventions/hiv_interventions.py
+++ b/stisim/interventions/hiv_interventions.py
@@ -7,7 +7,7 @@ import starsim as ss
 import numpy as np
 from stisim.interventions.base_interventions import STITest
 from stisim.interventions.utils import (
-    parse_coverage, _handle_deprecated_coverage, compute_coverage_target,
+    parse_coverage, compute_coverage_target,
 )
 from stisim.utils import count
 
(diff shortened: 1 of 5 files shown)


stisim/interventions/hiv_interventions.py
```diff
@@ -101,16 +101,16 @@ class ART(ss.Intervention):
...
-        - None: no coverage target; treat all who initiate (default)
-        - Scalar (e.g. 0.8): constant proportion of infected on ART
-        - Dict: {'year': [2000, 2020], 'value': [0, 0.9]} — linearly interpolated
-        - DataFrame: index=years, column 'n_art' (absolute numbers) or 'p_art'
-          (proportion of infected). Values are linearly interpolated to the sim
-          yearvec.
-        - Stratified DataFrame: columns Year, Gender/Sex, AgeBin (format
-          '[lo,hi)'), plus a numeric value column. Values are interpolated per
-          stratum.
+    Coverage is parsed by :func:`parse_coverage` and accepts:
```
cliffckerr: This is using Sphinx docstring style, in Quarto (and I think MkDocs, but might be switching to Quarto anyway, especially if this feature isn't available!) you can just refer to it directly: `parse_coverage` or `sti.parse_coverage()`.

stisim/interventions/hiv_interventions.py
```diff
@@ -101,16 +101,16 @@ class ART(ss.Intervention):
...
+        - ``None``: no coverage target; treat all who initiate (default)
```
cliffckerr, line 105: Likewise, just use single backticks

cliffckerr (approved):
Looks good though I did not go over all the details of the dataframe parsing!

PR #394, opened 2026-04-09
Examples gallery with first entry: ART interruptions

robynstuart:
## Summary
- New `docs/examples/` directory with Material card-based gallery index
- First example notebook: **Modeling ART interruptions** — builds scenario-specific coverage DataFrames, runs counterfactuals, compares outcomes
- Showcases mixed n/p coverage format from #383
- Gallery infrastructure: `pymdownx.emoji` for icons, CSS hover effects for cards

Depends on #389 (`mixed-coverage` branch)

### DIFF
--- a/docs/customization/css/extra.css
+++ b/docs/customization/css/extra.css
@@ -21,3 +21,15 @@ p:has(+ ul) {
   height: 2em;
   max-height: none;
 }
+
+/* Examples gallery cards */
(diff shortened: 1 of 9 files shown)


cliffckerr (approved):
Examples are a honking great idea -- let's do more of those! But seeing this also makes me want to switch to Quarto and use Markdown notebooks instead...

Issue #323, opened 2026-03-11
Implement and validate breastfeeding HIV transmission


robynstuart:
## Tasks

- [ ] Verify that HIV transmission via `MaternalNet` continues during breastfeeding period (postnatal network edges)
- [ ] Add parameterized breastfeeding transmission rate (currently uses the same beta as pregnancy transmission)
- [ ] Add result tracking for postnatal vs prenatal MTCT (see #325)

PR #401, opened 2026-04-09
Add MTCT tracking results to BaseSTI (#325)

### LINKED ISSUE #325: Easier MTCT tracking in results
Parent: #321

## Context

Currently MTCT infections can be computed as `new_infections - new_infections_sex` but this is not obvious and requires the user to know about both results. It would be better to have a dedicated result.

## Tasks

- [ ] Add `new_infections_mtct` result to the HIV module (count of infections where the source was via MaternalNet)
- [ ] Distinguish prenatal vs postnatal (breastfeeding) MTCT in results
- [ ] Add `n_on_art_pregnant` to default results (currently only added when pregnancy is in the sim, which is correct, but it should be documented)
- [ ] Add a simple test that verifies MTCT count is consistent: `new_infections_mtct + new_infections_sex == new_infections`
- [ ] Document how to access MTCT results in the user guide

robynstuart:
## Summary
- Add `new_infections_mtct` and `new_infections_sex` results to BaseSTI, available for all STI diseases (HIV, syphilis, etc.)
- Track MTCT count directly in `set_outcomes()` using the existing `age <= 0` check
- Simplify `MTCTransmissionCountTracker` test analyzer to use the new built-in result
- Add consistency test: `new_infections_sex + new_infections_mtct == new_infections`

Closes #325. Actually only closes 4/5 tasks in #325 but the last one is a separate issue (#323)

### DIFF
--- a/stisim/diseases/sti.py
+++ b/stisim/diseases/sti.py
@@ -116,8 +116,9 @@ class BaseSTI(ss.Infection):
         # Set initial prevalence
         self.init_prev_data = init_prev_data
 
-        # MTCT tracking: count of congenital infections this timestep, set in set_outcomes()
+        # Transmission route counts this timestep, set in set_outcomes()
         self._n_mtct = 0
+        self._n_sexual = 0
 
         # Results
         self.age_range = [15, 50]  # Age range for main results e.g. prevalence
@@ -284,20 +285,28 @@ class BaseSTI(ss.Infection):
         return new_cases, sources, networks
 
     def set_outcomes(self, uids, sources=None):
+        """
+        Set prognoses for newly infected agents and track transmission routes.
+
+        Args:
+            uids: UIDs of agents newly infected this timestep (not all agents)
+            sources: UIDs of the agents who transmitted to each new case
+        """
         super().set_outcomes(uids, sources=sources)
-        self.new_transmissions[:] = 0  # Total
-        self.new_transmissions_sex[:] = 0  # Sexual transmissions only
+        self.new_transmissions[:] = 0
+        self.new_transmissions_sex[:] = 0
         self._n_mtct = 0
+        self._n_sexual = 0
 
         if sources is not None:
-            # Separate sexual and congenital transmission
+            # Separate sexual and congenital transmission (age <= 0 = unborn)
             congenital = self.sim.people.age[uids] <= 0
-            self._n_mtct = np.count_nonzero(congenital)
+            self._n_mtct = ut.count(congenital)
+            self._n_sexual = ut.count(~congenital)
             sex_transmission = sources[~congenital]
             mtc_transmission = sources[congenital]
 
-            # Get the unique sources and counts
-            # Only needed for sexual transmission, not MTC.
+            # Track per-source transmission counts
             unique_sources, counts = np.unique(sources, return_counts=True)
             self.ti_transmitted[unique_sources] = self.ti
             unique_sources_sex, counts_sex = np.unique(sex_transmission, return_counts=True)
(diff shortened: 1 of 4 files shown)


stisim/diseases/sti.py
```diff
@@ -278,10 +287,12 @@ def set_outcomes(self, uids, sources=None):
         super().set_outcomes(uids, sources=sources)
         self.new_transmissions[:] = 0  # Total
         self.new_transmissions_sex[:] = 0  # Sexual transmissions only
+        self._n_mtct = 0
 
         if sources is not None:
```
ckirkman-IDM, line 301: I know this is not a changed line, but as far as I can tell, when set_outcomes is called, it can never be None.

infect() in class BaseSTI in sti.py always returns sources, which is has the appearance of never being None.  It never calls the diseases.py (superclass, ss.Infection) infect(), which even if it did, also looks like it never return None. Neither is commented/annotated in a way to indicate what a None would be even if it could be returned.

set_outcomes() is always passed the result of infect(), which would include the (non-None) sources. If I am correct in seeing that sources is never None, we should remove the strong suggestion in the code that it can be. Or if it CAN be None, we need to document in the function paths in infect(), set_ouutcomes(), and possibly ss.Infection.step().

stisim/diseases/sti.py
```diff
@@ -278,10 +287,12 @@ def set_outcomes(self, uids, sources=None):
         super().set_outcomes(uids, sources=sources)
```
ckirkman-IDM, line 295: There needs to be documentation that uids here must (only) represent new infections this ti. Otherwise the code below reads as setting all people <= 0 are being counted as mtct. The intent of the method is to only set some metadata/reporting related to new infections, and using it any other way will lead to miscounting.

The issue is compounded by a lack of documentation in the superclass set_outcomes() method (ss.Infection).

stisim/diseases/sti.py
```diff
@@ -278,10 +287,12 @@ def set_outcomes(self, uids, sources=None):
...
         self.new_transmissions[:] = 0  # Total
         self.new_transmissions_sex[:] = 0  # Sexual transmissions only
+        self._n_mtct = 0
 
         if sources is not None:
             # Separate sexual and congenital transmission
             congenital = self.sim.people.age[uids] <= 0
+            self._n_mtct = np.count_nonzero(congenital)
```
ckirkman-IDM: I suggest also directly recording sexual transmission onto self in the same way, for use later on in update_results() (more notes there).

stisim/diseases/sti.py
```diff
@@ -334,6 +345,11 @@ def update_results(self):
                     ask = f'{skk}_{ab1}_{ab2}'
                     self.results[f'{akey}{ask}'][ti] = ares[ai]
                     ai += 1
+
+        # Transmission route results
+        self.results['new_infections_mtct'][ti] = self._n_mtct
+        self.results['new_infections_sex'][ti] = self.results['new_infections'][ti] - self._n_mtct
+
```
ckirkman-IDM, line 361: Let's future proof this a little, as we know we will add breastfeeding transmissions. Please define (above) and set a value to self._n_sexual in set_outcomes() then use it here instead of the right-side of line 351. That way we can add breastfeeding (or any other category later) without updating the math in here.

tests/hiv_natural_history_analyzers.py
```diff
@@ -186,7 +186,7 @@ def update_results(self):
 class MTCTransmissionCountTracker(ss.Analyzer):
```
ckirkman-IDM, line 186: Do we even need this analyzer anymore (I know I made it :P)?

If not, let's just remove it and use default reporting only now.

ckirkman-IDM (changes requested):

cliffckerr (commented):
Did not review in detail but looks good to me

stisim/diseases/sti.py
robynstuart: https://github.com/starsimhub/stisim/issues/412

stisim/diseases/sti.py
robynstuart: Added

tests/hiv_natural_history_analyzers.py
robynstuart: Done in next PR

PR #408, opened 2026-04-10
Breastfeeding HIV transmission and prenatal/postnatal MTCT tracking (#323)

### LINKED ISSUE #323: Implement and validate breastfeeding HIV transmission

robynstuart:
## Summary
- Add `BreastfeedingNet` to hivsim.Sim default networks
- Add `beta_breastfeed` parameter to HIVPars (separate from prenatal `beta_m2c`)
- Replace `age <= 0` heuristic with network-type classification in `set_outcomes()`
- Add `new_infections_prenatal` and `new_infections_postnatal` results
- Combined `test_mtct` covering both routes with consistency checks

Depends on #401 (`add-mtct` branch)
Closes #323

### DIFF
--- a/stisim/diseases/sti.py
+++ b/stisim/diseases/sti.py
@@ -116,8 +116,9 @@ class BaseSTI(ss.Infection):
         # Set initial prevalence
         self.init_prev_data = init_prev_data
 
-        # MTCT tracking: count of congenital infections this timestep, set in set_outcomes()
-        self._n_mtct = 0
+        # MTCT tracking: counts set in set_outcomes()
+        self._n_prenatal = 0
+        self._n_postnatal = 0
 
         # Results
         self.age_range = [15, 50]  # Age range for main results e.g. prevalence
@@ -171,6 +172,8 @@ class BaseSTI(ss.Infection):
             betamap['structuredsexual'][1] = self.pars.beta_m2f * self.pars.rel_beta_f2m
         if self.pars.beta_m2c is not None and betamap and 'maternal' in betamap.keys():
             betamap['maternal'][0] = ss.permonth(self.pars.beta_m2c)
+        if hasattr(self.pars, 'beta_breastfeed') and betamap and 'breastfeeding' in betamap.keys():
+            betamap['breastfeeding'][0] = ss.permonth(self.pars.beta_breastfeed)
         if self.pars.beta_m2m is not None and betamap and 'msm' in betamap.keys():
             betamap['msm'][0] = self.pars.beta_m2m
             betamap['msm'][1] = self.pars.beta_m2m
@@ -236,11 +239,20 @@ class BaseSTI(ss.Infection):
         results += [
             ss.Result('new_infections_sex', dtype=int, label='New infections (sexual)', auto_plot=False),
             ss.Result('new_infections_mtct', dtype=int, label='New infections (MTCT)', auto_plot=False),
+            ss.Result('new_infections_prenatal', dtype=int, label='New infections (prenatal MTCT)', auto_plot=False),
+            ss.Result('new_infections_postnatal', dtype=int, label='New infections (postnatal MTCT)', auto_plot=False),
         ]
 
         self.define_results(*results)
 
         return
+
+    def step(self):
+        """ Override to pass network indices to set_outcomes for MTCT tracking """
+        new_cases, sources, networks = self.infect()
+        if len(new_cases):
+            self.set_outcomes(new_cases, sources, networks)
+        return new_cases, sources, networks
 
     def infect(self):
         """ Determine who gets infected on this timestep via transmission on the network """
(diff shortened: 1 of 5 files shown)


stisim/diseases/sti.py
```diff
@@ -347,8 +373,10 @@ def update_results(self):
                     ai += 1
 
         # Transmission route results
-        self.results['new_infections_mtct'][ti] = self._n_mtct
-        self.results['new_infections_sex'][ti] = self.results['new_infections'][ti] - self._n_mtct
+        self.results['new_infections_prenatal'][ti] = self._n_prenatal
+        self.results['new_infections_postnatal'][ti] = self._n_postnatal
+        self.results['new_infections_mtct'][ti] = self._n_prenatal + self._n_postnatal
+        self.results['new_infections_sex'][ti] = self.results['new_infections'][ti] - self.results['new_infections_mtct'][ti]
```
ckirkman-IDM: See other PR for math removal, but otherwise this looks good, but we need to clearly document that pre+post=mtct

tests/test_sim.py
```diff
@@ -81,9 +87,27 @@ def test_sim_creation():
     assert len(sim1.diseases) == 5, "Incorrect number of diseases initialized"
 
     # Test 2: mix of strings and modules
+    class MyNetwork(ss.Network):
```
ckirkman-IDM: What does this have to do with BF? Intentionally added or not?

tests/test_hiv.py
```diff
@@ -220,26 +220,40 @@ def test_doubling_hiv_sexual_beta_doubles_transmissions():
...
 
-    # setting fertility rate super high to enable shrinking the test agent/timewise
-    analyzer = MTCTransmissionCountTracker()
-    sim = build_testing_sim(analyzers=[analyzer], n_agents=500, duration=1, pregnancy=ss.Pregnancy(fertility_rate=1000))
+    # High fertility + high breastfeeding beta to ensure both transmission routes
+    sim = hivsim.demo('simple', run=False, plot=False, n_agents=2_000,
+                      beta_breastfeed=ss.permonth(0.1), init_prev=0.3)
     sim.run()
 
-    mtc_tranmissions = sim.results[analyzer.name][analyzer.result_name]
```
ckirkman-IDM, line 231: Oh good, unnecessary analyzer removed (noted in other PR).

stisim/diseases/sti.py
```diff
@@ -171,6 +172,8 @@ def validate_beta(self):
             betamap['structuredsexual'][1] = self.pars.beta_m2f * self.pars.rel_beta_f2m
         if self.pars.beta_m2c is not None and betamap and 'maternal' in betamap.keys():
             betamap['maternal'][0] = ss.permonth(self.pars.beta_m2c)
+        if hasattr(self.pars, 'beta_breastfeed') and betamap and 'breastfeeding' in betamap.keys():
```
ckirkman-IDM: I am assuming beta_breastfeed is a par if a BF network is added? Otherwise I don't see any other reference to it existing.

ckirkman-IDM: Also, betamap is always a dict, even if empty, so the "and betamap" is unnecessary given the final condition in the list.

ckirkman-IDM: Also, is there a reason to not mirror the style of beta_m2c just above this? If there isn't, the style difference implies there is a reason. But I don't think there is (I think both should use the same verification style for the par).

stisim/diseases/sti.py
ckirkman-IDM, line 180: I would love this method to have some documentation of what it is doing, how the beta pars/betamap are generated and mapped to each other. Otherwise it reads a bit like magic.

stisim/diseases/sti.py
```diff
@@ -283,21 +295,35 @@ def infect(self):
...
-            # Separate sexual and congenital transmission
-            congenital = self.sim.people.age[uids] <= 0
-            self._n_mtct = np.count_nonzero(congenital)
-            sex_transmission = sources[~congenital]
-            mtc_transmission = sources[congenital]
+            # Classify infections by network type (prenatal, postnatal, or sexual)
+            is_prenatal = np.zeros(len(uids), dtype=bool)
+            is_postnatal = np.zeros(len(uids), dtype=bool)
+            if networks is not None:
+                net_list = list(self.sim.networks.values())
```
ckirkman-IDM, line 334: Naming is confusing. Lines 309-313 is pretty unreadable and too obtuse without documentation and/or alternate naming.

- What does method parameter 'networks' mean, what is its type? I think 'networks' is really something like a list of 'network_indicies', but how they index isn't clear to me given that self.sim.networks looks to be a dict in the code.
- How is this different from 'self.sim.networks.values()'?

Without this information, it is extremely difficult to read the logic here to understand how the masking is being generated.

I do think, if I am correct, that more than one prenatal and postnatal networks can be utilized (hence |= them each time they pop up), which is a good call for adding.

ckirkman-IDM (changes requested):

cliffckerr (commented):
Looks fine to me, though tests are failing (says pregnancy module is missing)

stisim/diseases/sti.py
robynstuart: Done

tests/test_sim.py
robynstuart: Whoops! Removed

stisim/diseases/sti.py
robynstuart: Updated

PR #422, opened 2026-04-21
Remove BaseSTI.infect() override; tighten set_outcomes contract

### LINKED ISSUE #412: Check set_outcomes logic
"As far as I can tell, when set_outcomes is called, it can never be None." -- https://github.com/starsimhub/stisim/pull/401#discussion_r3088700854

### LINKED ISSUE #421: Reconsider infect() in BaseSTI
Do we need it or can we just rely on the parent class?

robynstuart:
## Summary

- Deletes `BaseSTI.infect()` — parent `ss.Infection.infect()` is a strict superset. The override duplicated the network-transmission loop but skipped the `ss.Route` (mixing pools) branch, so any non-Network route would `AttributeError` on `.edges`. Removing the override gains mixing-pool compatibility for free.
- Drops `sources=None` default and the `if sources is not None:` guard in `BaseSTI.set_outcomes()`. Per review on #401, `sources` is always passed by `step()` — the default and guard were dead.
- Keeps the `step()` override, which is still needed to thread `net_ids=networks` into `set_outcomes` for per-source MTC/sex transmission classification.
- Removes the now-unused `ss_int_` module-level alias.

Closes #412.
Closes #421.

## Test plan
- [x] Full test suite passes (`test_sim.py`, `test_stis.py`, `test_networks.py`, `test_hiv.py`, `test_connectors.py`, `test_hiv_interventions.py`, `test_baselines.py`, `test_calibration.py` — 57 tests)

### DIFF
--- a/stisim/diseases/sti.py
+++ b/stisim/diseases/sti.py
@@ -9,7 +9,6 @@ import sciris as sc
 import stisim.utils as ut
 import pandas as pd
 
-ss_int_ = ss.dtypes.int
 ss_float_ = ss.dtypes.float
 
 __all__ = ['BaseSTI', 'SEIS', 'BaseSTIPars', 'STIPars']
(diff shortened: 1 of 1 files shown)


ckirkman-IDM (approved):
Thanks!

Issue #420, opened 2026-04-21
Add an entry to the examples gallery


robynstuart:
Content TBC, but ideally something related to a new 1.5.4 feature

PR #423, opened 2026-04-22
Per-sex debut distributions + cleaner parameter routing in sti.Sim / hivsim.Sim

### LINKED ISSUE #397: AFS dynamics
Think of ways to get time-trends in age of debut

robynstuart:
## Summary

Originally a focused refactor of the `debut` pars, this PR grew to tackle the
broader parameter-routing issues surfaced in #425.

### 1. Debut distributions (closes #397, #420)

Replaces the shared `debut` / `debut_pars_f` / `debut_pars_m` setup in
`NetworkPars` with two per-sex distributions, so Starsim's native
callable-parameter support passes straight through:

```python
# Before
self.debut = ss.lognorm_ex(20, 3)
self.debut_pars_f = [20, 3]
self.debut_pars_m = [21, 3]
# set_debut then built per-agent mean/std arrays and overwrote the shared Dist

# After
self.debut_f = ss.lognorm_ex(20, 3)
self.debut_m = ss.lognorm_ex(21, 3)
# set_debut just calls rvs on each
```

Users can now do any of:
```python
sti.StructuredSexual(debut_f=18)                                   # scalar shortcut
sti.StructuredSexual(debut_f=ss.lognorm_ex(18, 2))                 # full Dist
sti.StructuredSexual(debut_f=ss.lognorm_ex(mean=fn_of_sim, std=3)) # callable
hivsim.demo('simple', debut_f=18)                                   # works via passthrough
```

New gallery entry at `docs/examples/dynamic_debut.ipynb` demonstrates
declining AFS over 1985-2025 by birth cohort (vs a static baseline).

### 2. Thin out hivsim.Sim

`hivsim.Sim` now sets `diseases='hiv'` as default (string form) and delegates
all HIV-pars routing to `sti.Sim.separate_pars`. Its own parameter-extraction
logic is gone. Defaults are exposed as `_default_demographics` /
`_default_networks` / `_default_interventions` for visibility.

The same override patterns now work uniformly across `hivsim.Sim`,
`hivsim.demo('simple')`, and `hivsim.demo('zimbabwe')`:
- `rel_trans_acute=26` (flat)
- `hiv=dict(rel_trans_acute=26)` (keyed dict)
- `hiv_pars=dict(rel_trans_acute=26)` (explicit kwarg)
- `debut_f=17` / `nw_pars=dict(debut_f=17)` (network)

The zimbabwe demo was also patched — previously it eagerly constructed HIV
with hardcoded calibrated pars and silently swallowed any user overrides
(the root cause of #425).

### 3. XOR validation + collision checks in sti.Sim

- `process_networks` / `process_stis` / `process_demographics` now raise
  `ValueError` if the user supplies both a module instance AND pars for the
  same slot. Previously user pars were silently swallowed.
- `separate_pars` raises on `pars ∩ kwargs` key collisions.
- The ambiguous-parameter warning now names which categories each ambiguous
  key lives in and suggests the explicit-kwarg fix.

### Breaking changes

- `NetworkPars.debut` (shared Dist) and `debut_pars_f` / `debut_pars_m` (lists)
  are gone. Use `debut_f` / `debut_m` instead, as scalar or `ss.lognorm_ex(...)`.
- Code that passed a module instance *and* pars targeting that module will
  now raise instead of silently ignoring the pars. E.g.
  `sti.Sim(diseases=[my_hiv], beta_m2f=0.03)` — choose one.

## Test plan

- [x] All 55 stisim tests pass (`test_sim`, `test_networks`, `test_hiv`,
      `test_hiv_interventions`, `test_calibration`, `test_stis`,
      `test_connectors`, `test_baselines`)
- [x] All tutorial + example notebooks execute end-to-end
      (`tut_intro`, `tut_interventions`, `tut_results`, `tut_cotransmission`,
      `tut_calibration`, `art_interruptions`, `dynamic_debut`)
- [x] XOR validation and `pars` ∩ `kwargs` collision verified manually to
      raise on intended mistakes and allow intended happy paths


tests/test_hiv_interventions.py
```diff
@@ -376,7 +376,10 @@ def test_art_parameter_sensitivity(do_plot=do_plot):
                 art = sti.ART(coverage=val)
             else:
                 art = sti.ART(coverage=0.5, art_initiation=ss.bernoulli(p=val))
-            sim = hivsim.demo('simple', run=False, plot=False, n_agents=n_agents, dur=10)
+            # Pin rand_seed=0: this test's signal is weak (coverage=0.5 caps total
+            # ART uptake regardless of art_initiation, making the lo vs hi margin
+            # within noise at n=1000). Pre-existing fragility, tracked separately.
+            sim = hivsim.demo('simple', run=False, plot=False, n_agents=n_agents, dur=10, rand_seed=0)
```
ckirkman-IDM, line 381: Does it make sense to have a coverage at all if sweeping over art_initiation? The documentation for class ART says:

"""
4. If no coverage is specified: all newly diagnosed who pass art_initiation go directly on ART (no capacity constraint)
""" 

I think this will yield a much better signal than default coverage 0.5 in the "else" block (sweeping art_initiation); it will also make the test make more sense, I think, than a hard-code on coverage for (some) sims in the loop.

stisim/networks.py
```diff
@@ -56,9 +56,8 @@ def __init__(self, **kwargs):
         )
```
ckirkman-IDM, line 56: You note in your PR summary that this change enables, "Starsim's native
callable-parameter support passes straight through:".

Would you mind showing me the logic trail that allows supporting the callable for ... oh nevermind, you/claude meant "parameters to ss.lognorm_ex()" can be callables now, got it (since the user can now directly pass the distribution to be used instead of it being networks.py-constructed from 2 params.

hivsim_examples/zimbabwe/sim.py
```diff
@@ -88,9 +88,21 @@ def make_sim(**kwargs):
...
     hiv_data = pd.read_csv(datadir / 'zimbabwe_hiv_data.csv')
 
-    # Create modules
-    hiv = sti.HIV(init_prev_data=init_prev, **sti_pars)
-    network = sti.StructuredSexual(condom_data=condom_data, **nw_pars)
+    # Pull user overrides for HIV / network out of kwargs so they reach module
+    # construction. Supports both disease-keyed dicts (`hiv=dict(...)`,
+    # `structuredsexual=dict(...)`) and flat kwargs. Matches sti.Sim test 3.
+    hiv_overrides = dict(kwargs.pop('hiv', None) or {})
+    for k in list(kwargs):
```
ckirkman-IDM, line 95: Small nit, but I find list(kwargs.keys()) to be more clear than simple list(kwargs).

tests/baseline.yaml
```diff
@@ -1,126 +1,130 @@
-cum_deaths: 4007371.0985
```
ckirkman-IDM: I am assuming this is essentially a regression test that gets updated if we think the differences are real/reasonable? Otherwise I am assuming this file is fine.

stisim/sim.py
```diff
@@ -127,16 +127,35 @@ def separate_pars(self, pars=None, sim_pars=None, sti_pars=None, nw_pars=None, d
             - If any key appears in both pars and *_pars, the value from *_pars will be used.
             - If any key appears in both pars and kwargs, the value from kwargs will be used.
         """
+        # Fail loudly on overlapping keys between `pars` and `**kwargs` — rather
```
ckirkman-IDM, line 130: I will admit that I am not super familiar with how pars works vs kwargs, but they do feel (and this code says so) overlapping in their functionality. I agree that if this is the case, then a user should definitely not be allowed to set up an ambiguous condition specifying a parameter twice in one call, as your update here addresses.

I would update this message to denote which key(s) overlap and what the value(s) are for each in pars/kwargs, e.g.

"""
Keys in both 'pars' and 'kwargs'. Specify each parameter in exactly one place:
paramA: pars: value1 kwargs: value2
paramB: pars: value 3 kwargs: value4
...
"""

Exact wording/structure can be up to you (above is just a quick suggestion), but it would be great for a user to see a listing of what values for which parameters are being passed. Could help them in a variety of ways to get things how they want it.

ckirkman-IDM:
On this point:

```
Breaking changes

...

Code that passed a module instance and pars targeting that module will
now raise instead of silently ignoring the pars. E.g.
sti.Sim(diseases=[my_hiv], beta_m2f=0.03) — choose one.
```

Do you mean something like the following instead? (pars vs kwargs):
```
sti.Sim(pars=HIVPars(beta_m2f=0.03), beta_m2f=0.03)
```

hivsim/sim.py
```diff
@@ -4,68 +4,98 @@
...
+    sti.Sim's ``separate_pars`` machinery. Users can customize:
+        - A whole module slot by passing the module directly (``diseases=MyHIV()``)
+        - Or by passing pars for the default module (``hiv=dict(rel_trans_acute=26)``
+          or flat ``rel_trans_acute=26``)
+        - But not both for the same slot (sti.Sim raises XOR violation).
+
+    Collisions between ``pars`` and ``kwargs`` raise an error: specify a value
+    exactly once.
     """
     def __init__(self, pars=None, sim_pars=None, hiv_pars=None, location=None, **kwargs):
```
ckirkman-IDM, line 49: This is getting confusing with pars, sim_pars, and hiv_pars, without documentation of what is in each and what is not in each.

ckirkman-IDM: For example, what is special about sim_pars vs pars? How would I as a user know what to pass?

ckirkman-IDM: Also, is it possible to remove the XOR checking of pars vs kwargs in hivsim.Sim? You put in stisim.Sim already, and it would be great to not be duplicative (perhaps update the super() call to stisim so _it_ can detect conflicts with the XOR?)

hivsim/sim.py
```diff
@@ -4,68 +4,98 @@
...
+            nw_kwargs = dict(pars.pop('nw_pars', None) or {})
+            for k in list(pars):
+                if k in sti.NetworkPars().keys():
+                    nw_kwargs[k] = pars.pop(k)
+            if nw_kwargs:
+                # Apply user's nw_pars to the default StructuredSexual.
+                pars['networks'] = [sti.StructuredSexual(**nw_kwargs), ss.MaternalNet()]
+                if any(isinstance(m, ss.Pregnancy) for m in sc.tolist(demographics)):
+                    pars['networks'].append(ss.BreastfeedingNet())
+            else:
```
ckirkman-IDM, line 89: Does this else clause need to be de-indented one level (on-level with "if 'networks' not in pars:")?

stisim/sim.py
```diff
@@ -127,16 +127,35 @@ def separate_pars(self, pars=None, sim_pars=None, sti_pars=None, nw_pars=None, d
...
+                    f'Specify each parameter in exactly one place.'
+                )
+
         # Merge in pars and kwargs
         all_pars = sc.mergedicts(pars, sim_pars, sti_pars, nw_pars, dem_pars, sim_kwargs, kwargs)
         all_pars = self.remap_pars(all_pars)  # Remap any parameter names
         vp, duplicates = self.get_valid_pars()
 
-        # Check if any parameter in all_pars is in duplicates
+        # Check if any parameter in all_pars is in duplicates. Duplicates are
```
ckirkman-IDM, line 147: Question: could there be legitimate overlapping parameter names? 

For example, dt. Would it be harmful for it to be passed to all matching modules?

Could there be others?

I know in other work I've done, I allowed duplicates in different module-equivalents, as it was a 'feature' that could be utilized to synchronize things between e.g. demographics and interventions (tied by a single, shared toggle) (not a feature I wanted to encourage without eyes-open, though).

stisim/sim.py
```diff
@@ -186,6 +205,13 @@ def separate_pars(self, pars=None, sim_pars=None, sti_pars=None, nw_pars=None, d
         self.nw_pars = nw_pars      # Parameters for the networks
         self.dem_pars = dem_pars
 
+        # Track just the user-provided contributions (separate from defaults)
+        # so process_* methods can raise XOR errors when the user supplies both
+        # a module instance and pars for the same slot.
+        self._user_sti_pars = dict(user_sti_pars)
```
ckirkman-IDM, line 213: Just verifying, this is to prevent a user from doing:

```
# not a pars/kwargs conflict, but module/kwargs?
sim = stisim.Sim(diseases=HIV(beta_m2f=0.01, beta_m2f=0.01)  
```

stisim/sim.py
```diff
@@ -262,7 +288,14 @@ def process_networks(self):
         If networks are provided, they will be used; otherwise, use default networks (usual case)
         """
         if len(self.pars['networks']):
```
ckirkman-IDM: Maybe just me, but adding " > 0" at the end of conditional network-length would be nice.

ckirkman-IDM (changes requested):
Thanks for helping to remove duplicative/colliding parameter inputs. I think the "required" changes should be pretty light, mostly questions + suggestions. I'll need to set some time aside soon to go over parameter handling in starsim, to make sure I know the full in and outs of their handling. This has helped, but it is apparently geared as a "it'll take almost anything as input" approach :D .

tests/test_hiv_interventions.py
robynstuart: updated

hivsim/sim.py
robynstuart: Added clarification to the docs

stisim/sim.py
robynstuart: Yes definitely, dt is the canonical example of a legitimately shared parameter. It appears in both SimPars (controls the base sim time step) and STIPars (controls each disease module's time step). When a user passes dt='month' as a flat kwarg, it's routed to SimPars first by our ordering, but this is intentional behaviour for most use cases, since you typically want all modules on the same time step. The warning is there to surface that silent routing is happening and invite users to be explicit if they need different values in different categories (e.g. sim_pars=dict(dt='month'), sti_pars=dict(dt='week')). We've left it as a warning rather than an error specifically for this reason.

hivsim/sim.py
robynstuart: Re the the XOR checking: the check in hivsim.Sim isn't a duplicate of the one in sti.Sim.separate_pars. hivsim.Sim merges pars + kwargs itself before calling super().__init__(), so by the time sti.Sim sees the call the original split is gone, it receives one merged dict and can't know whether a key appeared in both.

stisim/sim.py
robynstuart: updated

hivsim/sim.py
robynstuart: no this is right! I triple-checked

hivsim/sim.py
ckirkman-IDM: Is there a necessary reason why hivsim.Sim() is combining pars+kwargs before super() rather than passing them through to super? Playing devil's advocate a bit here, just to make sure we're not duplicating code.

PR #432, opened 2026-04-24
New prep intervention class with flexible targeting and supply constraint capability (Fix #330)

### LINKED ISSUE #330: Flexible PrEP targeting
Part of #329

## Summary

PrEP currently hardcodes FSW targeting in `step()`. The `eligibility` parameter exists but the default logic is baked into the class rather than being a default eligibility function.

## Current behavior

The eligibility check is embedded in `Prep.step()`:
```python
eligible = sim.networks.structuredsexual.fsw & ~hiv.infected & ~self.on_prep
```

The `eligibility` parameter is accepted but not cleanly integrated — it's unclear how it interacts with the built-in FSW filter.

## Proposed changes

1. Make the default eligibility an explicit function (e.g. `default_eligibility = lambda sim: sim.networks.structuredsexual.fsw & ~sim.diseases.hiv.infected`)
2. Allow users to override with any eligibility function, targeting any population
3. Separate the "not already on PrEP" and "not already infected" filters from the targeting filter — these should always apply regardless of eligibility
4. Document the default and how to override it

ckirkman-IDM:
I have made this a draft PR for the moment as I expect to add a few more tests and I am pretty sure there will be constructive feedback that I will want to incorporate before making a "cleaner" final approval PR.

Oh, and for fun, tests/test_product.py is claude code (after a couple of tries of providing feedback). Interesting in what it successfully inferred and the depth of how much it wants to test such a simple class! :D Realistically I'd remove some of the test code in it (feel free to chime ini), but I left it here at the moment for posterity/fun.


aakullian:
@ckirkman-IDM @robynstuart Here is a paper I came across that has a nice curve to PrEP discontinuation among AGYW and FSW in South Africa.  Could be a good starting point for parametrizing that component of the cascade.

https://pubmed.ncbi.nlm.nih.gov/36723509/

robynstuart:
Closing but will retain for future iterations

PR #439, opened 2026-04-30
Docs quarto

cliffckerr:
Switches docs build from MkDocs to Quarto.

Old home:
image

New home:
image

Old API (was never working, oops):
image

New API:
image

NB: I converted the tutorials from .ipynb to .qmd, but left the originals in there -- we can use either, both have their advantages.


cliffckerr:
The extra CSS is to enable Material-style cards:
image

JSchripsema-IDM:
@cliffckerr My only concern with leaving both qmd and ipynb files in the repo is them becoming out of sync. Are qmd files able to be connected to Binder or similar to run in the browser? I think that's a nice feature to get around the installation hurdle.

JSchripsema-IDM (approved):

cliffckerr:
> @cliffckerr My only concern with leaving both qmd and ipynb files in the repo is them becoming out of sync. Are qmd files able to be connected to Binder or similar to run in the browser? I think that's a nice feature to get around the installation hurdle.

Yes, I ended up removing the .ipynb files, since it turns out both JupyterLab and VS Code can handle the .qmd files natively (via JupyText).

For Binder, I haven't tried it out, but there is a way:
https://quarto.org/docs/projects/binder.html

You can also run on a GitHub Codespace, which might a better solution for many people -- Binder is incredibly slow I find.

robynstuart:
Looks great. Can you also rebase this to rc1.5.4 please? (If not too disruptive)

PR #441, opened 2026-05-01
Fix #330: make Prep eligibility explicit; decouple clinical filters from targeting

### LINKED ISSUE #330: Flexible PrEP targeting

robynstuart:
## Summary

- `Prep.step()` previously hardcoded `structuredsexual.fsw` regardless of the `eligibility` parameter, which was stored but never used.
- The FSW logic is now in an explicit `_default_eligibility` static method; `step()` always calls `self.eligibility(sim)`.
- `~hiv.infected` and `~self.on_prep` are applied unconditionally on top of whatever eligibility returns, so user-supplied callables only need to express targeting intent, not clinical preconditions.
- Default behaviour is unchanged (FSW targeting when no `eligibility` is passed).

## Test plan
- [ ] All 11 existing `test_hiv_interventions.py` tests pass
- [ ] Default (FSW) path smoke-tested
- [ ] Custom eligibility path smoke-tested (AGYW: `female & age < 25`)

Closes #330

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,3 +1,3 @@
-The unstaged changes show the file is in transition (seems to be resetting). Since the global `CLAUDE.md` already captures all essential context and the project's memory files document the ongoing work (birth key fixes, ANC screening, syphilis diagnosis), and no new gotchas have emerged:
+Memory is already documented. Since the CLAUDE.md file only has a one-liner and no substantive project context was added in the recent session (the git status shows just a CLAUDE.md edit), and the memory files already capture the active projects:
 
 SKIP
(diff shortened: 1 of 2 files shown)


cliffckerr (approved):
Looks good!

PR #443, opened 2026-05-01
Fix #369: upstream SyndromicMgmt from project repos into core

### LINKED ISSUE #369: Upstream SyndromicMgmt from project repos into core
Consolidate the two project-local SyndromicMgmt implementations (anc_sti_screening and stisim_vddx_zim) into core STIsim. The anc_sti_screening version is more mature. Relates to #345.

robynstuart:
## Summary

Consolidates the two project-local `SyndromicMgmt` implementations (`anc_sti_screening` and `stisim_vddx_zim`) into `stisim/interventions/base_interventions.py`. The `anc_sti_screening` version was used as the base (more mature).

**Changes vs. project versions:**
- `cervical_diseases` parameter replaces the hardcoded `ppl.ng.symptomatic | ppl.ct.symptomatic` check; defaults to any disease named `ng` or `ct` in the `diseases` list, so existing usage requiring no change
- Diagnostic accuracy (TP/FP/FN/TN) now derived from `cervical_diseases` membership rather than hardcoded `sim.diseases.ng/ct/tv` references
- Exported as `sti.SyndromicMgmt`

**Downstream repos** (`anc_sti_screening`, `stisim_vddx_zim`) can drop their local copies and import from stisim directly — that's a follow-up commit in each repo.

## Test plan
- [x] 24 existing tests pass (`test_stis.py`, `test_hiv_interventions.py`, `test_sim.py`)
- [x] `sti.SyndromicMgmt` importable from top-level

Closes #369

### DIFF
--- a/stisim/interventions/base_interventions.py
+++ b/stisim/interventions/base_interventions.py
@@ -11,7 +11,7 @@ from stisim.utils import count
 
 
 # %% Base classes
-__all__ = ["STIDx", "STITest", "SymptomaticTesting", "SyndromicMgmt", "STITreatment", "PartnerNotification", "ProductMix"]
+__all__ = ["STIDx", "STITest", "SymptomaticTesting", "SyndromicManagement", "STITreatment", "PartnerNotification", "ProductMix"]
 
 
 class STIDx(ss.Product):
@@ -424,7 +424,7 @@ class SymptomaticTesting(STITest):
         return
 
 
-class SyndromicMgmt(STITest):
+class SyndromicManagement(STITest):
     """
     Syndromic management of vaginal/urethral discharge syndromes.
 
@@ -442,24 +442,24 @@ class SyndromicMgmt(STITest):
     (``tx_mix_noncerv``), and males (cervical column of ``tx_mix_cerv``
     is reused for UDS).  Each dict maps outcome name → [female_prob, male_prob].
 
-    The ``outcome_treatment_map`` links each outcome to the treatment
-    module(s) that should be triggered.  Diagnostic accuracy (TP/FP/FN/TN)
-    is recorded on any disease module that exposes those result keys.
+    The ``outcome_tx_map`` links each outcome to the treatment module(s)
+    that should be triggered.  Diagnostic accuracy (TP/FP/FN/TN) is recorded
+    on any disease module that exposes those result keys.
 
     Args:
-        treatments (list):           treatment module instances
-        diseases (list):             disease modules to track accuracy for
-        cervical_diseases (list):    disease modules whose symptomatic flag
-                                     indicates cervical infection; defaults to
-                                     any disease named 'ng' or 'ct' in ``diseases``
-        outcome_treatment_map (dict): maps outcome name → list of treatment modules;
-                                     auto-constructed from ``treatments`` if omitted
-        treat_prob_data:             unused; reserved for future time-varying tx prob
+        treatments (list):        treatment module instances
+        diseases (list):          disease modules to track accuracy for
+        cervical_diseases (list): disease modules whose symptomatic flag
+                                  indicates cervical infection; defaults to
+                                  any disease named 'ng' or 'ct' in ``diseases``
+        outcome_tx_map (dict):    maps outcome name → list of treatment modules;
+                                  auto-constructed from ``treatments`` if omitted
+        treat_prob_data:          unused; reserved for future time-varying tx prob
         years, start, stop, eligibility, name, label: passed to STITest
 
     Examples::
 
-        syndromic = sti.SyndromicMgmt(
+        syndromic = sti.SyndromicManagement(
             diseases=[ng, ct, tv, bv],
             treatments=[ng_tx, ct_tx, metronidazole],
             eligibility=seeking_care_vds,
@@ -467,7 +467,7 @@ class SyndromicMgmt(STITest):
     """
 
     def __init__(self, pars=None, treatments=None, diseases=None,
-                 cervical_diseases=None, outcome_treatment_map=None,
+                 cervical_diseases=None, outcome_tx_map=None,
                  treat_prob_data=None, years=None, start=None, stop=None,
                  eligibility=None, name=None, label=None, **kwargs):
         super().__init__(years=years, start=start, stop=stop, eligibility=eligibility,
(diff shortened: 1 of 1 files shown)


stisim/interventions/base_interventions.py
```diff
@@ -424,6 +424,223 @@ def update_results(self):
         return
 
 
+class SyndromicMgmt(STITest):
```
cliffckerr: SyndromicManagement is a long name but it's also more consistent with SymptomaticTesting etc ...

stisim/interventions/base_interventions.py
```diff
@@ -424,6 +424,223 @@ def update_results(self):
...
+    """
+    Syndromic management of vaginal/urethral discharge syndromes.
+
+    When a symptomatic patient presents, they are probabilistically assigned
+    one of four treatment outcomes based on whether they have a cervical
+    infection (by default NG/CT) and their sex:
+
+        all3  — treated for cervical pathogens + TV/BV
```
cliffckerr, line 435: Don't love "all3", especially since only 2 drugs are being given! Maybe "all" or "both"? Which is most future proof, or unlikely to change since based on clinical guidelines?

stisim/interventions/base_interventions.py
```diff
@@ -424,6 +424,223 @@ def update_results(self):
...
+        self.pars.tx_noncerv_f.set(p=self.fvals_noncerv)
+        self.pars.tx_noncerv_m.set(p=self.mvals_noncerv)
+        # Resolve cervical diseases: use explicit list, else find 'ng'/'ct' in diseases
+        if not self._cervical_diseases:
+            self._cervical_diseases = [d for d in self.diseases if d.name in ('ng', 'ct')]
+        return
+
+    def init_results(self):
+        super().init_results()
+        results = sc.autolist()
```
cliffckerr: Since appending a list, can just use a regular list here

stisim/interventions/base_interventions.py
```diff
@@ -424,6 +424,223 @@ def update_results(self):
...
+        if sim.now >= self.start:
+            if uids is None:
+                uids = self.check_eligibility()
+                self.ti_tested[uids] = self.ti
+
+            if len(uids):
+                f_uids = uids[ppl.female[uids]]
+                m_uids = uids[ppl.male[uids]]
+
+                # Cervical infection: symptomatic in any cervical disease
```
cliffckerr: This block of code is somewhat complex logic, assigning a Boolean flag to a copy of an object looks odd to me

cliffckerr: Oh wait, is_cerv isn't a boolean, it's an array... maybe has_cerv_symp or cerv_symotomatic instead? Also don't love the custom initialization rather than using that as the default and updating it if the user supplies values

stisim/interventions/base_interventions.py
```diff
@@ -424,6 +424,223 @@ def update_results(self):
...
+                    all3=f_cerv_uids[ofc == 0] | f_noncerv_uids[ofnc == 0] | m_uids[om == 0],
+                    ngct=f_cerv_uids[ofc == 1] | f_noncerv_uids[ofnc == 1] | m_uids[om == 1],
+                    mtnz=f_cerv_uids[ofc == 2] | f_noncerv_uids[ofnc == 2] | m_uids[om == 2],
+                    none=f_cerv_uids[ofc == 3] | f_noncerv_uids[ofnc == 3] | m_uids[om == 3],
+                )
+
+                # Diagnostic accuracy: record TP/FP/FN/TN on each disease module
+                # all3 and ngct both treat cervical diseases; mtnz treats the rest
+                for disease in self.diseases:
+                    if not hasattr(disease, 'sex_keys'):
```
cliffckerr, line 590: Would prefer to avoid hasattr checking if possible. What determines whether or not this attr exists? Can we use isinstance instead?

stisim/interventions/base_interventions.py
```diff
@@ -424,6 +424,223 @@ def update_results(self):
...
+                        disease.results[f'new_false_pos{skk}'][self.ti] += len(outcomes['all3'] & disease.susceptible & ppl[pattr])
+                        disease.results[f'new_true_neg{skk}'][self.ti]  += len(outcomes['none'] & disease.susceptible & ppl[pattr])
+                        disease.results[f'new_false_neg{skk}'][self.ti] += len(outcomes['none'] & disease.treatable & ppl[pattr])
+                        disease.results[f'new_true_pos{skk}'][self.ti]  += len(treat_outcomes & disease.treatable & ppl[pattr])
+                        disease.results[f'new_false_pos{skk}'][self.ti] += len(treat_outcomes & disease.susceptible & ppl[pattr])
+                        disease.results[f'new_false_neg{skk}'][self.ti] += len(miss_outcomes & disease.treatable & ppl[pattr])
+                        disease.results[f'new_true_neg{skk}'][self.ti]  += len(miss_outcomes & disease.susceptible & ppl[pattr])
+
+                # Route to treatments
+                for outcome, txs in self.outcome_treatment_map.items():
```
cliffckerr: For consistency with other naming this could be outcome_tx_map instead

cliffckerr (commented):
Looks good, some minor refactoring comments. Should this be v1.6.0 instead of 1.5.4?

stisim/interventions/base_interventions.py
robynstuart: 3 drugs are given! Unlikely to change

stisim/interventions/base_interventions.py
robynstuart: changed

stisim/interventions/base_interventions.py
robynstuart: have opened an issue: #446

stisim/interventions/base_interventions.py
robynstuart: Renamed to has_cerv_symp and refactored as suggested - empty BoolArr is the default

stisim/interventions/base_interventions.py
robynstuart: done

stisim/interventions/base_interventions.py
robynstuart: done

PR #444, opened 2026-05-01
Fix #370: add ANCTest + InfantHIVTest; remove project-specific ANCScreen from core

### LINKED ISSUE #370: Upstream ANCScreen from anc_sti_screening
Move the GA-windowed, multi-disease ANCScreen intervention from anc_sti_screening into core STIsim for reuse across country models.

### LINKED ISSUE #406: Compare model MTCT rates against published estimates
e.g. ~15-45% without intervention, <5% with full PMTCT

robynstuart:
### `ANCTest` + `InfantHIVTest` (closes #370)

`ANCTest` models a single ANC visit per pregnancy scheduled at a random gestational timepoint. `visit_prob` controls attendance. At the visit, diseases are auto-detected from the sim via a built-in registry:

- **HIV** — positives set `hiv.diagnosed` / `hiv.ti_diagnosed`; ART links up automatically
- **Syphilis** — positives route to a `disease_treatment_map` entry

Both diseases optionally schedule a newborn/infant test on the unborn via `maternalnet` + `pregnancy.ti_delivery` when a mother tests positive. Additional diseases can be added via `disease_names` + `disease_treatment_map`. Per-disease `test_sensitivity` is configurable; all defaults are overridable.

`InfantHIVTest` is a thin `HIVTest` subclass that fires only at its scheduled timestep. Scheduled by `ANCTest` or by `HIVTest` directly (new `newborn_test=` parameter) using the same `maternalnet` mechanism as `NewbornSyphTest`.

### `BreastfeedingNet` removed from `sti.Sim` defaults (closes #406)

`BreastfeedingNet` is only relevant for diseases with postnatal transmission (HIV, syphilis). Removed from `sti.Sim` default networks — sims that need it should add it explicitly.

### `Prep` legacy parameter handling

`years=` was previously extracted from `self.pars` after `update_pars`, which required it to survive as a recognised par. It is now popped from `kwargs` before `update_pars` runs, making the legacy conversion explicit and avoiding reliance on internal par state.

### Test cleanup

Removed fragile `len(sim.diseases/networks/interventions) == N` assertions from `test_sim.py`. These break whenever a default module is added or removed, and test count rather than content.

## Test plan
- [x] 24 tests pass
- [x] `sti.ANCTest` and `sti.InfantHIVTest` importable


stisim/interventions/base_interventions.py
```diff
@@ -641,6 +641,183 @@ def store_results(self):
...
+        hiv      — sets ``hiv.diagnosed`` / ``hiv.ti_diagnosed``; optionally schedules
+                   ``InfantHIVTest`` on the unborn via ``maternalnet``
+        syphilis — routes positives to ``disease_treatment_map['syphilis']``; optionally
+                   schedules ``NewbornSyphTest`` on the unborn via ``maternalnet``
+    Additional diseases in ``disease_names`` are tested and routed via
+    ``disease_treatment_map`` with configurable sensitivity.
+
+    Newborn/infant tests must be added to ``sim.interventions`` by the caller and
+    passed via ``newborn_tests`` so they go through normal Starsim initialisation.
+    ``ANCTest`` only handles scheduling them (sets their ``ti_scheduled`` state).
```
cliffckerr: This explanation feels like it might be a bit more complicated than it needs to be

stisim/interventions/base_interventions.py
```diff
@@ -641,6 +641,183 @@ def store_results(self):
...
+
+    Args:
+        visit_prob (float):           probability of attending ANC (default 0.85)
+        disease_names (list):         diseases to test for; default: auto-detect
+                                      'hiv' and 'syphilis' from sim
+        test_sensitivity (dict):      per-disease sensitivity, e.g. ``{'hiv': 0.99}``;
+                                      default 1.0 for all
+        disease_treatment_map (dict): maps disease name → treatment intervention;
```
cliffckerr, line 663: Do we need both `disease_names` and `disease_treatment_map`? Something feels not quite right to me about the API, I don't know if it's just the different names (`disease_names`, `disease_treatment_map`, `newborn_tests`) or if there would be a fundamentally different way of handling this. Seems like there could be some better parallels between mothers/newborns and tests/treatments ...

stisim/interventions/base_interventions.py
```diff
@@ -641,6 +641,183 @@ def store_results(self):
...
+    _registry = {
+        'hiv':      dict(use_diagnosed=True),
+        'syphilis': dict(use_diagnosed=False),
+    }
+
+    def __init__(self, visit_prob=0.85, disease_names=None, test_sensitivity=None,
+                 disease_treatment_map=None, newborn_tests=None,
+                 start=None, stop=None, name=None, label=None, **kwargs):
+        super().__init__(name=name, label=label, **kwargs)
+        self._visit_prob_val  = visit_prob
```
cliffckerr: Why underscores and not just store them directly?

stisim/interventions/base_interventions.py
```diff
@@ -641,6 +641,183 @@ def store_results(self):
...
+
+        if sim.now < self.start or sim.now >= self.stop:
+            return
+
+        # Schedule a visit for each newly pregnant woman
+        if hasattr(sim.demographics, 'pregnancy'):
+            newly_preg = (sim.demographics.pregnancy.ti_pregnant == ti).uids
+            attendees  = self._visit_prob_dist.filter(newly_preg)
+            months_dt  = ss.dur(1, 'month') / sim.t.dt   # timesteps per month
+            offsets    = np.round(self._visit_timing.rvs(attendees) * months_dt).astype(int)
```
cliffckerr: Can use `dist.randround()` to get CRN-safe randomly rounded values if that's important

stisim/interventions/base_interventions.py
```diff
@@ -641,6 +641,183 @@ def store_results(self):
...
+
+        visit_uids = (self.ti_visit == ti).uids
+        if not len(visit_uids):
+            return
+
+        self.results['n_attended'][ti] = len(visit_uids)
+
+        for dname in self._active_diseases:
+            if dname not in sim.diseases:
+                continue
```
cliffckerr: Will this silently fail if a disease is mistyped, eg if you do `syph` when it's supposed to be `syphilis`?

stisim/interventions/base_interventions.py
```diff
@@ -641,6 +641,183 @@ def store_results(self):
...
+                self._sens_dist.set(p=sens)
+                pos_uids = infected[self._sens_dist.rvs(infected)]
+            else:
+                pos_uids = infected
+
+            if not len(pos_uids):
+                continue
+
+            self.results[f'n_{dname}_positive'][ti] += len(pos_uids)
+            cfg = self._registry.get(dname, {})
```
cliffckerr: I don't love this ... feels very CS!

stisim/interventions/base_interventions.py
```diff
@@ -641,6 +641,183 @@ def store_results(self):
...
+            if dname in self._newborn_tests:
+                self._schedule_newborn(sim, pos_uids, self._newborn_tests[dname])
+        return
+
+    @staticmethod
+    def _schedule_newborn(sim, pos_mother_uids, newborn_test):
+        """Schedule a newborn/infant test for unborn children of positive mothers."""
+        if not (hasattr(sim.networks, 'maternalnet') and
+                hasattr(sim.demographics, 'pregnancy')):
+            return
```
cliffckerr: It should be `if not ('maternalnet' in sim.networks)` etc. But the logic is fragile, since it will break if someone renames the network or subclasses `MaternalNet`. (Same with pregnancy.)

Would prefer for maybe a `sim.get_maternalnet()` method that finds maternl nets based on type, and errors if there is >1, and this method also errors (rather than silently fails) if it's looking for one and can't find it.

stisim/interventions/hiv_interventions.py
```diff
@@ -87,21 +87,58 @@ class HIVTest(STITest):
...
     def step(self, uids=None):
         sim = self.sim
         outcomes = super().step(uids=uids)
         pos_uids = outcomes['positive']
         sim.diseases.hiv.diagnosed[pos_uids] = True
         sim.diseases.hiv.ti_diagnosed[pos_uids] = self.ti
+
+        # Schedule infant HIV test for unborn children of newly diagnosed mothers
+        if self.newborn_test is not None and len(pos_uids):
+            if hasattr(sim.networks, 'maternalnet') and hasattr(sim.demographics, 'pregnancy'):
```
cliffckerr: Same comment as previous file

stisim/interventions/hiv_interventions.py
```diff
@@ -87,21 +87,58 @@ class HIVTest(STITest):
...
+    A positive result sets ``hiv.diagnosed`` on the infant, enabling ART linkage.
+
+    Args:
+        test_prob (float): probability of the infant receiving the test once
+                           scheduled (default 1.0 — if scheduled, it happens)
+        name, label:       standard
+    """
+    def __init__(self, test_prob=1.0, name=None, label=None, **kwargs):
+        super().__init__(name=name, label=label, **kwargs)
+        self._test_prob_dist = ss.bernoulli(p=test_prob)
```
cliffckerr: Would prefer without the underscore prefix

tests/baseline.yaml
```diff
@@ -1,91 +1,92 @@
-cum_deaths: 3807751.1185
-deaths_cmr: 0.012810721938616333
+breastfeedingnet_n_edges: 326246.1395124481
```
cliffckerr, line 1: Why did the baseline change?

cliffckerr (commented):
Looks pretty good, although there's some overly cautious, vanilla Python code (e.g. `if hasattr(sim.networks, 'maternalnet'): ... else continue`) that I'm worried is both fragile and could obfuscate user mistakes. Would prefer more opinionated named methods instead (e.g. `mnet = sim.get_maternalnet(); if mnet is None: raise Exception('Cannot schedule newborn tests with no maternal network')`.

stisim/interventions/base_interventions.py
robynstuart: Trimmed in 48f4a6b — collapsed the registry/auto-detect explanation into two short paragraphs.

stisim/interventions/base_interventions.py
robynstuart: Keeping the three flat args (`disease_names`, `disease_treatment_map`, `newborn_tests`) — they let callers override one piece without restating the others, and auto-detect (`['hiv', 'syphilis']`) covers the common case so most users only touch `disease_treatment_map`. A single nested-dict `disease_config` would be tidier but more verbose for the typical call site.

stisim/interventions/base_interventions.py
robynstuart: Dropped in 48f4a6b — args are now stored under their input names (`self.visit_prob`, `self.disease_names`, …).

stisim/interventions/base_interventions.py
robynstuart: Switched to `self.visit_timing.randround(...)` in 48f4a6b for CRN-safe rounding.

stisim/interventions/base_interventions.py
robynstuart: Now raises `KeyError` from `init_pre` listing valid disease names (48f4a6b), so mistyped `'syph'` etc. fails fast instead of silently skipping.

stisim/interventions/base_interventions.py
robynstuart: Replaced the `_registry` dict + `cfg.get('use_diagnosed')` dispatch with an explicit `if dname == 'hiv': … elif dname in self.disease_treatment_map: …` (48f4a6b). Auto-detect is now a plain `['hiv', 'syphilis']` class attr.

stisim/interventions/base_interventions.py
robynstuart: Replaced both `hasattr` paths with `sim.get_module(ss.MaternalNet, die=False)` + `sim.get_module(ss.Pregnancy, die=False)` (48f4a6b). If either is missing when newborn tests are scheduled, raises `RuntimeError` with a clear message — no more silent no-op, and robust to subclassing/renaming.

stisim/interventions/hiv_interventions.py
robynstuart: Same fix applied here in 48f4a6b — `sim.get_module(ss.MaternalNet, die=False)` / `sim.get_module(ss.Pregnancy, die=False)` with explicit `RuntimeError` if either is missing.

stisim/interventions/hiv_interventions.py
robynstuart: Dropped in 48f4a6b.

tests/baseline.yaml
robynstuart: `BreastfeedingNet` was removed from `sti.Sim` default networks (issue #406 — only relevant for diseases with postnatal transmission, like HIV/syphilis). The baseline reflects that change in default network composition.

robynstuart:
Thanks for the review @cliffckerr — addressed in 48f4a6b:

- **Underscore prefixes** dropped throughout (`visit_prob`, `disease_names`, `disease_treatment_map`, `newborn_tests`, all `*_dist` fields).
- **`hasattr` checks** replaced with `sim.get_module(ss.MaternalNet, die=False)` + `sim.get_module(ss.Pregnancy, die=False)`. Raises `RuntimeError` when either is missing instead of silently no-op'ing.
- **Mistyped disease names** now raise `KeyError` from `init_pre` listing valid names.
- **Registry dispatch** replaced with an explicit `if dname == 'hiv' … elif …` branch.
- **`randround`** for CRN-safe rounding of visit timing.
- **Docstring** trimmed.

26 unit + 23 HIV tests pass. PR is now `MERGEABLE` (rc1.5.4 merge in 386bcef).

PR #452, opened 2026-05-05
Add sti.route_pars; refactor separate_pars (#451)

### LINKED ISSUE #451: Rethink separate_pars
Can it be moved upstream so we can call it from the hivsim demos?

robynstuart:
- New `sti.route_pars` utility (in `stisim/utils.py`) that auto-discovers par categories (sim, sti, nw, dem) from the par-class registry and routes flat kwargs accordingly. Cross-category clashes (e.g. `dt` valid in both sim and sti) broadcast to all matches with a printed note; unknown keys (typos) raise when `strict=True`.
- `sti.Sim.separate_pars` shrinks from ~95 to ~30 lines by delegating routing to `route_pars`. Keeps only sti.Sim-specific concerns: legacy renames (`beta` → `beta_m2f`), nested per-module dicts (`hiv=dict(...)`), defaults, and the `_user_*_pars` stash for downstream XOR checks. Drops the `pars`-vs-`kwargs` collision raise (last-wins now) and the redundant `get_valid_pars` helper.
- hivsim demo builders (`simple`, `zimbabwe`) call `route_pars` to consume network pars from kwargs before constructing modules. Flat-kwargs API restored: `hivsim.demo('simple', debut_f=18)` and `hivsim.demo('zimbabwe', beta_m2f=0.04, prop_f0=0.7)` now work.
- `docs/examples/dynamic_debut.qmd` reverted to the flat-kwargs form the prose advertises.

Closes #451.

### DIFF
--- a/docs/examples/dynamic_debut.qmd
+++ b/docs/examples/dynamic_debut.qmd
@@ -102,7 +102,7 @@ msim = ss.parallel(sim_static, sim_dynamic)
 
 def afs_by_cohort(sim, cohort_bin=5):
     ppl = sim.people
-    net = sim.networks.structuredsexual
+    net = sim.networks.mfnetwork
     # .values gives numpy arrays for active agents only
     debut  = net.debut.values
     age    = ppl.age.values
--- a/hivsim_examples/simple/sim.py
+++ b/hivsim_examples/simple/sim.py
@@ -13,12 +13,19 @@ import hivsim
 
 
 def make_sim(**kwargs):
-    """Create a simple HIV simulation using ``sti.MFNetwork``."""
+    """Create a simple HIV simulation using ``sti.MFNetwork``.
+
+    Flat ``**kwargs`` are auto-routed via ``sti.route_pars`` — pass network
+    pars (``debut_f=18``, ``prop_f0=0.7``), sim pars (``n_agents=5000``,
+    ``dur=10``), or sti pars (``init_prev=0.05``) directly.
+    """
     kwargs.setdefault('n_agents', 2000)
     kwargs.setdefault('dur', 20)
+    nw_kwargs = sti.route_pars(kwargs).nw
+    for k in nw_kwargs: kwargs.pop(k)  # consumed here, don't re-route in sti.Sim
 
     if 'networks' not in kwargs:
-        nets = [sti.MFNetwork(), ss.MaternalNet()]
+        nets = [sti.MFNetwork(**nw_kwargs), ss.MaternalNet()]
         # BreastfeedingNet only makes sense if Pregnancy is in demographics.
         # User-supplied demographics: respect them. Otherwise hivsim.Sim defaults
         # to Pregnancy + Deaths, so include BreastfeedingNet.
@@ -26,6 +33,5 @@ def make_sim(**kwargs):
         if demog is None or any(isinstance(m, ss.Pregnancy) for m in sc.tolist(demog)):
             nets.append(ss.BreastfeedingNet())
         kwargs['networks'] = nets
-
     kwargs.setdefault('interventions', [sti.HIVTest(), sti.ART(), sti.VMMC()])
     return hivsim.Sim(**kwargs)
--- a/hivsim_examples/zimbabwe/sim.py
+++ b/hivsim_examples/zimbabwe/sim.py
@@ -17,15 +17,15 @@ import stisim as sti
 datadir = sc.thispath()
 
 # Sim settings
-sim_pars = dict(start=1990, stop=2025, age_scale=1000)
+default_sim_pars = dict(start=1990, stop=2025, age_scale=1000)
 
 # Calibrated parameters
-sti_pars = dict(
+default_hiv_pars = dict(
     beta_m2f=0.035,
     eff_condom=0.95,
     rel_init_prev=1.0,
 )
-nw_pars = dict(
+default_nw_pars = dict(
     prop_f0=0.79,
     prop_m0=0.83,
     f1_conc=0.16,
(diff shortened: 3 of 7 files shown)


ckirkman-IDM:
I generated a clade review for this and read through it; raw text for it is below. I don't think all of it is super critical, so here is my take on the issues it identified, 1-7:

1. I wouldn't bother adding caching unless we find that it is needed for performance improvement. Caching can have unexpected consequences later on that can make bugs hard to track. Avoid premature optimization :) .

2. Interesting, but most likely the solution at some point is to formalize the requirement of exposed .pars . The (None, None) issue is basically the same as in the PR I reviewed yesterday. Taking a closer look (e.g. at hiv_syph connector in hiv_sti.py), and I see the culprit:

```
def __init__(self, hiv_module, syphilis_module, pars=None, **kwargs):
```

The REAL issue is that there is no way to ask the class, hiv_syph in this case, what its default parameters are. Let us make a separate issue to examine this, as I do think it would be much cleaner if each class had a way to ask it what parameters exist and what the default values are *without* instantiation. It would also help standardize all the modules, connectors + others.

3. Please update the doc string.

4. I saw this on a visual pass on the code before trying claude. If there is a valid reason for omitting connector_pars in the way suggested, I'm ok with it. Otherwise, I'd make it mirror the other types of pars.

5.  Claude is basically asking if nw_pars could be used in other *_pars catetories (e.g., other modules). I believe that we are currently NOT allowing this per discussions last week; at least by default. Let me know if I am mistaken. If broadcasting IS allowed (by toggle somehow), we may need to make sure that information is made available here so nw_pars can be broadcastable. I'll defer to your memory + expertise in knowing if there is an issue here or not. (possibly documentation regarding said broadcasting; allowed? not? sometimes?)

6. Perhaps add verbose to the signature of separate_pars() and route the value to sti.route_pars() call (line ~131 stisim/sim.py)

7. Unittests would be nice for route_pars(), especially since no other tests are updated/touched by this PR.


---
## Code Review: Add `sti.route_pars`; refactor `separate_pars`

The refactor is correct and the public API is clean. A few issues worth addressing before merge:

---

### Issues

**1. `_par_registry()` is rebuilt on every call — no caching**

Every call to `route_pars()` rebuilds the entire registry, instantiating every STI module, network, demographic module, and all 5 connectors. With the example builders calling `sti.route_pars(kwargs)` and then `sti.Sim(...)` calling `separate_pars()` → `route_pars()` again, this runs at minimum twice per sim construction. A `functools.lru_cache` with no args (or a simple module-level sentinel) would fix this entirely.

---

**2. `merged_connector_pars()` relies on an implicit, unenforced constructor contract**

```python
def merged_connector_pars():
    merged = {}
    for cls in connector_register().values():
        merged.update(cls(None, None).pars)  # relies on (None, None) being valid
    return merged
```

This requires every registered connector to accept `(None, None)` and expose a usable `.pars`. The docstring documents the convention but doesn't enforce it. If a future connector validates its inputs in `__init__`, this silently breaks the registry for all callers. A `cls.default_pars()` classmethod would make the contract explicit and enforceable.

---

**3. `route_pars` docstring omits the `connector` return key**

```
Returns:
    sc.objdict with keys ``sim``, ``sti``, ``nw``, ``dem``, ``unmatched``.
```

The actual return has six keys — `connector` is missing. Callers reading the docstring won't know the bucket exists.

---

**4. No `connector_pars` named argument despite `connector` being a first-class bucket**

`route_pars` accepts `sim_pars`, `sti_pars`, `nw_pars`, `dem_pars` as named pre-categorized inputs, but there's no `connector_pars` equivalent. The asymmetry will be noticed once users try to pre-supply connector pars the same way.

---

**5. Latent double-routing bug in example builders**

```python
# hivsim_examples/simple/sim.py
nw_kwargs = sti.route_pars(kwargs).nw
for k in nw_kwargs: kwargs.pop(k)  # consumed here, don't re-route in sti.Sim
```

If a network par is also valid in another category (the broadcast case), `route_pars()` includes it in `nw_kwargs`, it gets popped from `kwargs`, and it never reaches the second category in `sti.Sim`. The broadcast path is exactly the case where manual pop-and-re-route is most error-prone. `zimbabwe/sim.py` has the same pattern.

---

**6. `verbose=True` default uses bare `print()`**

Library code using `print()` can't be filtered by log level, redirected, or silenced programmatically (except by passing `verbose=False`, which callers don't know to do). `warnings.warn(..., stacklevel=2)` or `sc.printv` would be more appropriate. As-is, any cross-category broadcast during `sti.Sim(...)` construction prints to stdout with no way for the user to suppress it.

---

**7. No unit tests for `route_pars`**

Coverage exists only via `Sim` construction. None of the following are directly tested:
- Broadcast behavior (key valid in 2+ categories)
- `strict=True` raising with suggestions
- The `connector` bucket being populated
- Pre-categorized `sim_pars`/`sti_pars`/etc. arguments
- `unmatched` dict under `strict=False`

Given this is now public API (`__all__`) and called directly by example builders, it warrants its own test cases.

---

### Minor

- `separate_pars()` docstring doesn't mention the `connector` category or the new `self._user_connector_pars` attribute.
- The nested-dict detection heuristic in `separate_pars()` (`any(gk in sti_keys for gk in v)`) is fine in practice but worth a comment explaining why the pre-scan is necessary before calling `route_pars`.

---

### Summary

| Area | Status |
|---|---|
| Correctness | Correct as written; latent double-routing risk in example builders |
| Design | Clean abstraction; `connector_pars` omission from signature is an asymmetry |
| Performance | `_par_registry()` rebuild per call is the main concern |
| Tests | No direct unit tests for `route_pars` |
| Docs | `connector` bucket missing from docstring return type |

The two items most worth fixing before merge: **`_par_registry()` caching** (affects every sim construction) and **the missing `connector` in the docstring/signature**. The double-routing risk in the example builders could be a follow-up issue if not addressed here.

robynstuart:
Thanks @ckirkman-IDM!
1. Agree, let's not cache
2. Made an issue here: https://github.com/starsimhub/starsim/issues/1335
3. Done
4. Done
5. ... (see below)
6. done
7. added in tests/test_parameters, deliberately minimal but can build up as needed

Point 5 is the big one, have gone back and forward on this a LOT. As I see it, there are three distinct cases and only the third is the "implicit broadcast" one I'd be cautious about:
- Universal pars like dt. Every module has these. Starsim already broadcasts: sti.Sim(dt=ss.years(0.5)) cascades to every module under it. To override per-module you do sti.Sim(dt=ss.years(0.5), diseases=sti.HIV(dt=ss.month)). Nothing new in this PR.
- Multi-disease shared pars like beta_m2f, init_prev. These are deliberately registered with the same name across STI disease modules -  that part of the point of parameters.py and the registry. sti.Sim(diseases=['hiv','syph'], beta_m2f=0.04) setting both is a feature: the same one-knob convenience that exists for dt, just at a different level. Per-disease overrides are explicit: sti_pars=dict(hiv=dict(beta_m2f=0.05), syph=dict(beta_m2f=0.4)) or diseases=[sti.HIV(beta_m2f=...), sti.Syph(beta_m2f=...)].
- Unintended clashes introduced by user-supplied custom modules. We can't statically prevent these, a custom module can re-use any name. Today route_pars prints a one-line note (suppressible via verbose=False) listing the categories the key landed in. Allowed, not an error, and now documented in the route_pars docstring under "Design: when does broadcasting happen?".

So the answer to "is broadcasting allowed?" is: cases 1 and 2 are by design, and disabling them would either error or silently drop legitimate one-knob usage; case 3 is broadcast-with-warning, which I think is the right balance for the custom-module escape hatch. The new docstring lays this out but let me know if you want me to surface any of it more prominently (e.g. in sti.Sim's class docstring).

PR #453, opened 2026-05-05
Auto-add coinfection connectors; route flat connector pars (#442)

### LINKED ISSUE #442: Automatic connectors not working
sim = sti.Sim(diseases=['hiv', 'syph']) should automatically add an hiv-syphilis connector.

robynstuart:
**Auto-add connectors.** `sti.Sim(diseases=['hiv', 'syph'])` now auto-instantiates `sti.hiv_syph` (and any other registered `<d1>_<d2>` connector for the disease pairs in the sim). Lookup tries `sti.<d1>_<d2>` then `sti.<d2>_<d1>` (reversing the constructor arg order to match). Pairs with no registered class are silently skipped. Replaces the `NotImplementedError`-on-`connectors=True` stub.

**Flat connector pars.** `sti.Sim(diseases=['hiv','syph'], rel_sus_hiv_syph=3.5)` routes the par into the auto-added `hiv_syph` connector. Each auto-added connector receives only the kwargs it accepts. Pars that match no built connector raise (catches typos and config errors like passing `rel_sus_hiv_syph` without syph in the disease list).

Skipped if the user supplies their own connectors via `connectors=[...]` — pass an explicit list to take full control. Mixing flat connector pars with `connectors=[...]` raises rather than silently dropping the flat pars.

**Cleanup.** Removes `sti.Sim.case_insensitive_getattr` — no callers across stisim, hivsim, or any localization repo.

Closes #442.

### DIFF
--- a/stisim/parameters.py
+++ b/stisim/parameters.py
@@ -6,7 +6,8 @@ import starsim as ss
 import stisim as sti
 
 
-__all__ = ['SimPars', 'sti_aliases', 'sti_register', 'merged_sti_pars', 'make_sti', 'mergepars', 'dem_pars']
+__all__ = ['SimPars', 'sti_aliases', 'sti_register', 'merged_sti_pars', 'make_sti', 'mergepars',
+           'dem_pars', 'connector_register', 'merged_connector_pars']
 
 
 class SimPars(ss.SimPars):
@@ -130,6 +131,28 @@ def mergepars(*args):
     dicts = [dict(sc.dcp(arg)) for arg in args if arg is not None]
     merged_pars = sc.objdict(sc.mergedicts(*dicts))
     return merged_pars
+
+
+def connector_register():
+    """Auto-discoverable connectors keyed by ``<d1>_<d2>``. Only includes
+    connectors that follow the ``(disease1_module, disease2_module, ...)`` constructor
+    convention, so ``sti.Sim`` can build them automatically from a disease list."""
+    from stisim.connectors.hiv_sti import hiv_syph, hiv_tv, hiv_ng, hiv_ct, hiv_bv
+    return sc.objdict(
+        hiv_syph=hiv_syph,
+        hiv_tv=hiv_tv,
+        hiv_ng=hiv_ng,
+        hiv_ct=hiv_ct,
+        hiv_bv=hiv_bv,
+    )
+
+
+def merged_connector_pars():
+    """Merge all parameters from auto-discoverable connectors."""
+    merged = {}
+    for cls in connector_register().values():
+        merged.update(cls(None, None).pars)
+    return merged
 
 
 # Demographic pars
--- a/stisim/sim.py
+++ b/stisim/sim.py
@@ -143,6 +143,7 @@ class Sim(ss.Sim):
         self._user_sti_pars = {**routed.sti, **nested_sti}
         self._user_nw_pars  = dict(routed.nw)
         self._user_dem_pars = dict(routed.dem)
+        self._user_connector_pars = dict(routed.connector)
 
         return routed.sim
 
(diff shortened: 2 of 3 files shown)


stisim/parameters.py
```diff
@@ -132,6 +133,28 @@ def mergepars(*args):
...
+        hiv_ct=hiv_ct,
+        hiv_bv=hiv_bv,
+    )
+
+
+def merged_connector_pars():
+    """Merge all parameters from auto-discoverable connectors."""
+    merged = {}
+    for cls in connector_register().values():
+        merged.update(cls(None, None).pars)
```
ckirkman-IDM, line 154: cls(None, None) seems weird here (and each other location in this PR). I think simply cls() (no arguments should) do it and would be a bit cleaner.


However, it still seems odd that instantiation is needed to grab pars with their default values, when querying HIVPars directly should do it. However, I'll consider this detail out of scope.

stisim/utils.py
```diff
@@ -21,10 +21,11 @@ def _par_registry():
...
     return {
-        'sim': set(sti.SimPars()),
-        'sti': set(sti.merged_sti_pars()),
-        'nw':  set(sti.NetworkPars()),
-        'dem': set(sti.dem_pars()),
+        'sim':       set(sti.SimPars()),
+        'sti':       set(sti.merged_sti_pars()),
+        'nw':        set(sti.NetworkPars()),
+        'dem':       set(sti.dem_pars()),
+        'connector': set(sti.merged_connector_pars()),
```
ckirkman-IDM, line 28: If a user supplies their own connector, e.g., for hiv and syph, would **its** pars be caught here and reported? Or the pars of the default connector? This is my main issue with the code; it either cannot read user-created connector parameters or it isn't very clear that it can and the code should better document this potential (given the auto-loading addition, as a user, I'd want to make sure there is no cross-over between *my* connector and the default one.

stisim/sim.py
```diff
@@ -142,6 +143,7 @@ def separate_pars(self, pars=None, sim_pars=None, sti_pars=None, nw_pars=None, d
         self._user_sti_pars = {**routed.sti, **nested_sti}
         self._user_nw_pars  = dict(routed.nw)
         self._user_dem_pars = dict(routed.dem)
+        self._user_connector_pars = dict(routed.connector)
```
ckirkman-IDM, line 146: looking at sti.route_pars(), routed.connector is always a {}? I'm not sure how recording a {} here is recording user supplied connector pars. If I'm missing something, it's too simple or obtuse.

Update: Ok, I found it, pars and kwargs (assuming the merging of them works) given to route_pars() can route to 'connector'.

ckirkman-IDM (changes requested):
One major point to check on, the rest is minor.

stisim/utils.py
robynstuart: Good question, had to think about this one :) 

In the general case, Starsim modules allow you to pass any module par as a kwarg, since all kwargs get passed to the module's update_pars method. STIsim threads this logic up to the sim constructor, with automatic routing to put module pars into their correct module. But only for the out-of-the box modules, NOT for user-supplied ones - the registry that resolves kwargs is built from the default out-of-the-box module classes only.  

I made a deliberate design choice not to allow users to mixing custom modules with kwargs; it's too error-prone and we can't ensure that there won't be unintentional name clashes, plus if a user if going to all the trouble of defining their own module then I think it's fair to assume they'll be comfortable doing e.g. `connectors=[MyConn(rel_sus_hiv_syph=3.5)`

In any case, I've now tightened the docstrings on process_connectors and _par_registry to make this explicit, and to describe the process for custom-connector users to follow.

stisim/parameters.py
robynstuart: Using cls() alone raises TypeError: missing 2 required positional arguments because both disease modules are required positional. The (None, None) is the minimum that lets __init__ run far enough to populate self.pars.

PR #457, opened 2026-05-06
Replace PartnerNotification with PriorPartners-based implementation

robynstuart:
- Replace `PartnerNotification` with a `PriorPartners`-based implementation with two channels (current sexual network + optional prior-partner recall) and separate notification × attendance probabilities per channel
- The previous transmission-graph-based version is removed (it was a stub with no internal users)
- New gallery example `docs/examples/partner_notification.qmd`
- Tests in `test_hiv_interventions.py` verify end-to-end behaviour, attendance ≤ notification, and PN-vs-baseline diagnosis difference

## Why
Built for the `sti_notification` undertreatment analysis (companion to `syph_dx_zim`), but the upgraded class is generally useful: it lets STIsim users model partner notification as a real-world program (notify → attend → test) using the active sexual network and a recall window, rather than relying on the disease transmission graph.

## API change
- Old (removed): `PartnerNotification(eligibility, test, test_prob=0.5)` — single probability, traversed `disease.log`
- New: `PartnerNotification(eligibility, test, p_notify_current, p_attends_current, p_notify_previous, p_attends_previous, current_network, previous_network)` — all are `ss.bernoulli` distributions; set `p_notify_previous=ss.bernoulli(p=0)` to disable the prior-partner channel


ckirkman-IDM (approved):
LGTM

PR #463, opened 2026-05-12
Fix ART stratified coverage to allocate per-(age,sex) stratum

aakullian:
Stratified ART coverage data was being parsed correctly but then collapsed to a single aggregate target before correction, with `art_coverage_correction` allocating ART globally by CD4 priority across the whole population. This washed out the age/sex differentials in the input: older/sicker agents were over-treated at the expense of younger groups, and male/female gaps were smoothed away. The model could not reproduce its own stratified inputs.

Changes:
- Add `compute_stratum_targets()` in `interventions/utils.py`: parallel to `compute_coverage_target` but returns the per-stratum dict instead of the summed total. Returns None for absent/aggregate coverage.
- `ART._get_n_to_treat` now returns `(total, stratum_targets)`.
- `ART.art_coverage_correction` dispatches: stratified targets get per-stratum correction; aggregate targets keep the existing global behaviour.
- New `_art_coverage_correction_stratified` adjusts each (age_bin, sex) stratum independently using the same CD4/care-seeking priority rules as the global path, applied within-stratum.
- Strengthen `test_art_stratified_coverage_matching` to assert per-stratum matching (previously only printed). Tolerance is generous to accommodate finite-population noise and the diagnosis lag from HIVTest.

### DIFF
--- a/stisim/interventions/hiv_interventions.py
+++ b/stisim/interventions/hiv_interventions.py
@@ -7,7 +7,7 @@ import starsim as ss
 import numpy as np
 from stisim.interventions.base_interventions import STITest
 from stisim.interventions.utils import (
-    parse_coverage, compute_coverage_target,
+    parse_coverage, compute_coverage_target, compute_stratum_targets, age_sex_mask,
 )
 from stisim.utils import count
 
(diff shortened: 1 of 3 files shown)


robynstuart:
Thanks @aakullian, this looks great!!

I pushed a small fix to test_art_stratified_coverage_matching. Two things were going on:

1. The test built coverage data for 2005/2015 but the sim only runs 2000–2005, so the assertion was comparing measured coverage against the 2015 target the sim never reaches. Switched to 2000/2005.
2. At n_agents=1k split across 6 (age × sex) strata, individual strata had 1–2 infected agents which was too noisy for the test. Dropped to 4 strata (15–30 and 30+ for each sex).

I also flattened the coverage to a single constant target across both data years, so the assertion window doesn't catch the ramp-up. All 4 strata now land within tol=0.20. If this test ends up frgaile over future releases we might need to move it to devtests and/or think of another way to test this is working as expected, but it's looking ok for now.

PR #464, opened 2026-05-14
Add dur_dx2tx on HIVTest for delay from diagnosis to ART start (#398)

### LINKED ISSUE #398: Time delay from HIV+ test to ART initiation
See EMOD dynamics for ref, discuss with Adam/Daniel for realistic dynamics

robynstuart:
## Summary
- Adds `dur_dx2tx` to `HIVTest` (default `ss.constant(0)`) for a per-pathway delay between diagnosis and ART start.
- `HIVTest` schedules `hiv.ti_art = ti + delay` for positive results; `ART` becomes passive and starts agents whose `ti_art == ti & ~on_art`. Coverage pool uses `ti_art <= ti & ~on_art`.
- `ANCTest` and initially-diagnosed cases (`init_post`) schedule with no delay — backward compatible.
- `new_agents_on_art` result is gated on `on_art` (since `ti_art` may now hold a future scheduled time).

Closes #398.

## Test plan
- [x] `tests/test_hiv_interventions.py` — all 13 tests pass, including new `test_art_dx2tx_delay` verifying:
  - With `dur_dx2tx=ss.constant(ss.years(2))`, no agents on ART before the delay elapses.
  - `ti_art == ti_diagnosed + delay` for on-ART agents.
  - Default (`ss.constant(0)`) is backward compatible (`ti_art == ti_diagnosed`, agents on ART from ti=0).
- [x] Full repo test suite (66 tests) passes.

Usage:
```python
sti.HIVTest(test_prob_data=0.3, dur_dx2tx=ss.constant(ss.years(0.5)))  # 6-month delay
```

### DIFF
--- a/stisim/diseases/hiv.py
+++ b/stisim/diseases/hiv.py
@@ -511,8 +511,7 @@ class HIV(BaseSTI):
         self.results['cum_deaths'][ti] = np.sum(self.results['new_deaths'][:ti + 1])
         self.results['new_diagnoses'][ti] = np.count_nonzero(self.ti_diagnosed == ti)
         self.results['cum_diagnoses'][ti] = np.sum(self.results['new_diagnoses'][:ti + 1])
-        # ti_art may also hold a scheduled future start; gate on on_art for actual starts
-        self.results['new_agents_on_art'][ti] = sum((self.ti_art == ti) & self.on_art)
+        self.results['new_agents_on_art'][ti] = sum((self.ti_art == ti) & self.on_art)  # gate on on_art; ti_art may also hold a scheduled future start
         if self.include_mtct:
             pregnant = self.sim.people.pregnancy.pregnant
             self.results['n_on_art_pregnant'][ti] = np.count_nonzero(self.on_art & pregnant)
--- a/stisim/diseases/hiv.py
+++ b/stisim/diseases/hiv.py
@@ -211,6 +211,8 @@ class HIV(BaseSTI):
         initial_cases_diagnosed = self.pars.init_diagnosed.filter(initial_cases)
         self.diagnosed[initial_cases_diagnosed] = True
         self.ti_diagnosed[initial_cases_diagnosed] = 0
+        # Schedule ART start at ti=0 (no delay) so an ART intervention picks them up immediately
+        self.ti_art[initial_cases_diagnosed] = 0
         return
 
     # CD4 functions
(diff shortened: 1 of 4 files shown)


stisim/diseases/hiv.py
```diff
@@ -509,7 +511,8 @@ def update_results(self):
         self.results['cum_deaths'][ti] = np.sum(self.results['new_deaths'][:ti + 1])
         self.results['new_diagnoses'][ti] = np.count_nonzero(self.ti_diagnosed == ti)
         self.results['cum_diagnoses'][ti] = np.sum(self.results['new_diagnoses'][:ti + 1])
-        self.results['new_agents_on_art'][ti] = np.count_nonzero(self.ti_art == ti)
+        # ti_art may also hold a scheduled future start; gate on on_art for actual starts
+        self.results['new_agents_on_art'][ti] = sum((self.ti_art == ti) & self.on_art)
```
cliffckerr: Would prefer for the line comment to be on the same line to not break up the flow

stisim/interventions/hiv_interventions.py
```diff
@@ -90,18 +94,36 @@ class HIVTest(STITest):
...
             self.eligibility = lambda sim: ~sim.diseases.hiv.diagnosed
+
+        # Newborn test
         self.newborn_test = newborn_test
+        
+        # Deal with pars
+        self.define_pars(
+            dur_dx2tx=ss.constant(0),
+        )
+        self.update_pars(pars, **kwargs)
```
cliffckerr, line 123: A few things that could be improved here:
- The docstring seems out of date with the function signature (at minimum the arguments are in a different order)
- Different parameters are handled differently -- there's no default for most parameters; there are defaults for product and dur_dx2tx but these two are handled separately

Not sure if this would work technically, but would suggest something like:
```py
_ = None # Convention for when values are replaced with defaults
    def __init__(self, product=_, test_prob_data=_, years=_, start=_,
                 eligibility=_, name=_, label=_, newborn_test=_, **kwargs):

        super().__init__(test_prob_data=test_prob_data, years=years, 
                         start=start, eligibility=eligibility, name=name, label=label) # this stays the same

        # Deal with pars
        self.define_pars(
            product=HIVDx(name=f'HIVDx_{name}'),
            newborn_test=<something>,
            dur_dx2tx=ss.constant(0),
        )
        self.update_pars(**kwargs)
```

Things to check would be:
1) is it OK to call `super().__init__()` without the product and does it get initialized correctly in the next step
2) seems like this would name the product `'HIVDX_'` by default, which is probably not ideal ... maybe `self.name` would fix that if `super().__init__()` has already been called?

cliffckerr (commented):
Looks good, suggested a different architecture for initializing the intervention, would need testing though

stisim/diseases/hiv.py
robynstuart: done

stisim/interventions/hiv_interventions.py
robynstuart: Yes good idea. `define_pars(product=HIVDx(...))` works as a default, but user override via `product=MyDx(...)` fails with `TypeError: Cannot update a module with <class 'MyDx'>: must be a dict to set new parameters.` Leaving for now but we can open an issue

PR #477, opened 2026-05-21
New MFNetwork match_pairs() algorithm

### LINKED ISSUE #459: MFNetwork (possibly others) does not properly statistically sample partnerships to preserve desired age distributions.
In the following section of method MFNetwork.match_pairs(), there is no statistical sampling to ensure that each "limited count" gendered agent draws a partner following the female's normal-distributed desired age(mean, sd). Instead, the algorithm globally selects N women and N men, then lines them up in order of low<->high age (men) and low<->high rvs desired age. This results in females selecting men in order of the female age preferences, each taking the youngest remaining man. This *ignores* the female's preference, in that there could be a glut of too-young (less desirable) men that they will erroneously select in the current algorithm. It would be better for preserving the parameterized age gap (and female preferences to be able to skip over the less desirable, younger men and straight to older, more desirable men. This would boost selected-male ages, which would fix the current error of "resultant parameterized age gap is too small (male-female).

I am not 100% sure what the **best** fix is yet (currently thinking on it), but the analysis here is almost certainly the cause of the bug.

https://github.com/starsimhub/stisim/blob/0f3bb6f05aafc28c029b557ca927865344de9f05/stisim/networks.py#L537-L544

ckirkman-IDM:
Algorithm fixes the negative/downward and growing bias in modeled age gap at relationship formation time. The current algorithm is now within < 5% (still a slight downward bias) and hits the targeted stddevs well, too.

Algorithm has been optimized for speed about as much as it can be by batch-binary searching of the male ages the females want as their "top initial pick".

Algorithm description:
- females are ordered by their desired_age which is age + normal(target_gap, stddev) , drawn at each timestep, for females who are underpartnered.
- underpartnered males are ordered by age.
- females pick, in order of lowest to highest desired_age, the male whose age is closest to desired_age who has **not already** been picked in this timestep. Any match found must have a (desired_age - age < 1) to help eliminate spurious, large-gap matches.
- After matches are made, any with an actual (age_gap > target_gap + 3*stddev) are removed. target_gap and stddev here are the **largest** values in the age_diff_pars grid for speed.
- In order to get a close match to requested age gaps for older men/women, it was necessary to implement a mechanism by which women who **might** pick men over 55 (default value, can be changed) have a progressively reduced chance to look for a partner, even if they are underpartnered. This reduces the downward bias that is introduced when women in this demographics try to pick men of an age where none are available (upper end of normal dist. tail), but those in their cohort who pick a lower age can almost always find a match. The current mechanism is located on line 553 currently in networks.py, and does:

```
actual_looking_chance = ((AGE_TAPER_CUT - f_age) / TAPER_OFFSET_Y)  ** 1.5
# note: any value outside 0-1 before raising to the power is trimmed to 0-1
```
AGE_TAPER_CUT is a param in MFParams and is the max age at which females can form new relationships
TAPER_OFFSET_Y is the number of years before AGE_TAPER_CUT at which female chances to **actually** seek a new relationship starts diminishing. It is auto-set based on age_diff_pars.

This effectively reduces the oversupply of female demand (relative to older male supply) which strongly reduces downward age_gap bias.

Happy to discuss!

Relevant test files:
tests/test_mfnetwork.py -- actual asserts on age gaps
tests/test_network_diagnostics.py -- graphs (png files) and tables relating to run time, age gaps, etc.


aakullian (commented):
Thanks @ckirkman-IDM, fix worked really well.

I ran the new algorithm against a diagnostic sweep (`age_diff_pars` μ ∈ {0, 7, 14}, σ=3, full Eswatini-model sim) and the three configurations cleanly separate where the old algorithm collapsed them. Comparison plot attached.

Two things worth discussing before merge:

### 1. Realized μ falls ~1 year below configured μ

Population-weighted across all pairings: configured μ=7 → realized μ=6.2. The under-shoot is concentrated in the youngest male bin (man 15-24: realized gap +4.0 vs configured 7), and held across every `max_deviation` value and every σ I tried (including σ=5-6 with DHS-realistic pars).

I think this is structural: male concurrency + long stable-partnership durations leaves an age-skewed available pool, and closest-male matching picks "younger than desired" because that's where availability concentrates. Not something the algorithm can fix, but worth documenting: a calibration that trusts configured μ as the realized μ will mis-attribute the missing year to other parameters.

Suggest the docstring add a "Limitations" note that `age_diff_pars` is not orthogonal to concurrency/`stable_dur_pars` in a calibration setting.

### 2. `max_deviation` has no effect on age-pairing

md=1, md=3, and md=5 produce realized means within 0.05 of each other across every age bin. The 3σ post-hoc trim does all the tail shaping. Either drop the parameter or document that it is only useful when populations are small and partners are limited.

Image

---

Overall: clean fix to a calibration-blocking bug. Happy to share the diagnostic harness and DHS Eswatini reference data if useful for `tests/test_mfnetwork.py`.

ckirkman-IDM:
@aakullian 
Thanks for the review notes; I updated the comments/notes in the code to reflect your thoughts. There is a regression test failing currently saying that the changes are significant enough to stop and think. I'll wait for Robyn to chime in (she'll be back next week) to merge, but I think these results make sense. Below, sim2 is the "new" test simulation and sim1 is the pre-new-algorithm. Honoring the age gaps better means that higher-prevalence men (older) are matching more with younger females, which could plausibly lead to more HIV infections, more HIV infections of pregnant women, and in the presence of ART, more women pregnant on ART. There is quite a lot of more women pregnant on ART! There must be some detail of the targeting or the ART implementation that these women are "prone" to being eligible for.
 
image

PR #492, opened 2026-06-05
Logistics

### LINKED ISSUE #493: Create initial logistics module
It will contain the ported SuppliedIntervention, Supplies, Supply, Product classes from the existing PrEP redesign. These classes are not dependent upon PrEP in any way, but are potentially useful for many STI and/or starsim projects. Moving them into stisim/logistics is a starting place, and their current feature set is a starting point.

The only additional work is to discuss and decide how to merge ss.Product and logistics Product, as there is no known reason why we need separate product feature sets at the moment. This issue will either merge the two in an appropriate way or create a follow-on issue to resolve their differences for future work.

ckirkman-IDM:
**Not** fully inspected from claude enhancements/normalization, one of the reasons this is a _draft_ for now.

The other reason is that ss.Product is not yet merged with Product (used in logistics). I am posting claude's set of options for "fixing this" in addition to the code, so the code can be looked at alongside the options.


ckirkman-IDM:
# Merging `logistics.Product` and `ss.Product` — options analysis

Analysis for the standing TODO on `stisim/logistics/product.py`:
`# TODO: determine subclassing potential with ss.Product`. This documents what each
class is, why merging is harder than a rename, and the candidate approaches with
trade-offs. **No decision is made here.**

## What the two classes actually are

| | `ss.Product` (+ `Dx`/`Tx`/`Vx`) | `logistics.Product` |
|---|---|---|
| **Base** | `ss.Module` — sim-integrated (lifecycle `init_pre`, results, states, name, RNG) | plain object (POJO) |
| **Purpose** | the *mechanism of effect* — acts on agents | a *supply-chain / economic descriptor* — passive data |
| **Core method** | `administer(uids)` → applies effect, stochastic, disease-aware | none; just `efficacy_at_ti(ti)` lookup |
| **Efficacy model** | per-administration draw keyed on **disease + health_state** (a `df` + `ss.choice`/`ss.bernoulli`) | **`eff_by_ti`**: a durability curve over time-since-delivery |
| **Metadata** | `name`, `df`, `diseases`, `health_states`, `hierarchy` | `name`, `category` (enum), `delivery_mode` (enum), `cost` |
| **Validation / RNG** | reproducible `ss` distributions; minimal input validation | strict enum + range validation; no RNG |
| **Lives in** | **upstream starsim** (`starsim/products.py`) | this repo (`stisim/logistics`) |
| **Used by** | starsim core; **stisim** `STIDx`, `HIVDx`, `ProductMix` (which use `administer(sim, uids)`) | only `Supply`/`Supplies`/`SuppliedIntervention` + tests |

**The headline:** these aren't two implementations of the same thing — they're two
*layers*. `ss.Product` is "how a product changes an agent." `logistics.Product` is
"what the product is, what it costs, and how its protection decays after delivery."
They overlap only on `name` and the word "efficacy."

## Three things that make this harder than a class rename

1. **The efficacy models genuinely conflict.** `ss.Tx` = one Bernoulli per
   administration, keyed by disease/health-state. `logistics.Product` = a time-indexed
   durability curve. Merging the *classes* does not merge these; you must either keep
   both (each product uses the relevant one) or design a unified efficacy abstraction.
   **This is the real crux — bigger than the hierarchy question.**
2. **`ss.Product` is upstream.** Editing it is a *starsim* change (coordination +
   release), not a stisim-local one. Options that stay in stisim avoid that.
3. **`ss.Product` is an `ss.Module`.** It expects sim lifecycle/name/RNG and is
   typically added to a sim. `logistics.Product` is a passive object freely constructed
   inside `Supply` and in tests. Any unification that makes the descriptor a Module
   changes how `Supply` builds/holds it (and the tests).

## Options

### Option 1 — Push logistics fields *up* into `ss.Product` (the suspected path)
Add `category`/`delivery_mode`/`cost`/`eff_by_ti` (+validation) to `ss.Product` as
*optional* fields; retire `logistics.Product`; `Supply` consumes `ss.Product`.
- **Pros:** one class; every product (incl. `Dx`/`Tx`/`Vx`) gains cost/category/
  durability; logistics aligns with the sim-integrated product.
- **Cons:** upstream starsim edit + coordination; bloats the base with fields many
  products don't need (a `Dx` has no meaningful `delivery_mode`); the two efficacy
  models still coexist unresolved; existing `STIDx`/`HIVDx`/`ProductMix` and
  `Dx`/`Tx`/`Vx` must keep working (so fields must default to `None`).

### Option 2 — `logistics.Product` *subclasses* `ss.Product`
`class Product(ss.Product)` adding the logistics fields; one lineage, stays in stisim.
- **Pros:** logistics product becomes a real, administer-able sim product carrying
  logistics metadata; reuses Module machinery; no upstream change.
- **Cons:** it becomes an `ss.Module` — heavier construction (lifecycle, name, often
  needs a sim), which is overkill for something `Supply` uses as passive data;
  tests/`Supply` would need to build a Module; `administer` must be implemented or left
  abstract; efficacy models still coexist.

### Option 3 — Composition: `logistics.Product` *links to* an `ss.Product`
Keep the logistics descriptor; give it an optional `effect: ss.Product`. Durability/
cost/category live in logistics; the agent-effect delegates to the linked
`ss.Product.administer(...)` (e.g. from `SuppliedIntervention`'s `dist_func`).
- **Pros:** clean separation of concerns; no upstream change; no forced Module-ness;
  each layer keeps the efficacy model that fits it.
- **Cons:** two objects per product, not the "single class" goal; some conceptual
  duplication.

### Option 4 — A `LogisticsMixin` on the `ss.Product` family
Put `category`/`delivery_mode`/`cost`/`eff_by_ti`+validation in a mixin; apply it where
logistics matters (`class SuppliableTx(ss.Tx, LogisticsMixin)`); `Supply` keys off the
mixin/attributes; retire `logistics.Product`.
- **Pros:** `ss.Product` hierarchy untouched and unbloated; logistics behavior is
  opt-in; one object per product.
- **Cons:** mixin/MRO cooperation with `ss.Module.__init__` needs care; still must
  reconcile efficacy models; "single class" only per concrete combination.

### Option 5 — Shared `Protocol`/interface, two classes
Define a `SuppliableProduct` Protocol (the attributes `Supply` needs: `name`, `cost`,
`category`, `efficacy_at_ti`). Both `logistics.Product` and adapted `ss.Product`
subclasses satisfy it; `Supply` depends on the protocol.
- **Pros:** smallest change; decouples `Supply` from a concrete class; enables gradual
  convergence.
- **Cons:** explicitly *not* a single class; duck-typing rather than unification.

## How to frame the decision

- If the goal is literally **one class** and starsim can be changed: **Option 1** (with
  all logistics fields optional) is the cleanest end-state, but commits you to an
  upstream change and to resolving the efficacy-model question.
- If you want unification **without touching upstream**: **Option 2** (subclass) or
  **Option 4** (mixin). Mixin avoids forcing Module-ness onto purely-logistics products;
  subclass is simpler but heavier.
- If, on reflection, these are better as **two layers**: **Option 3** (compose) or
  **Option 5** (protocol) are lower-risk and arguably more honest about the separation —
  at the cost of not being "one class."

**The first decision isn't the hierarchy — it's the efficacy model.** Decide whether
`eff_by_ti` (time-since-delivery durability) and the `df`/Bernoulli per-administration
model should become one abstraction or remain two; that choice constrains which of the
options above is even coherent.

## Behaviors in `ss.Product` worth preserving in any path
- `administer(uids)` — the effect mechanism.
- `ss.Module` sim-integration — lifecycle, results, reproducible RNG via `ss`
  distributions.
- disease/health-state-aware outcomes (`df`).

These are exactly what `logistics.Product` lacks and what real interventions rely on, so
whatever path is chosen, the merged/bridged class should keep them.

## Source references
- `ss.Product` / `Dx` / `Tx` / `Vx`: `starsim/products.py` (upstream).
- `logistics.Product`: `stisim/logistics/product.py`.
- stisim subclasses of `ss.Product`: `STIDx`, `ProductMix`
  (`stisim/interventions/base_interventions.py`), `HIVDx`
  (`stisim/interventions/hiv_interventions.py`) — these call `administer(sim, uids)`.

ckirkman-IDM:
Despite what claude thinks, I believe there should ideally be just one efficacy model, whatever we choose that to be, so to me options 1 & 2 seem the most natural, as there should really be one Product source for everyone, not separated concerns etc (these are the same concerns; what is a Product?) given as reasons for the other options.

I think one natural pattern would be that the SuppliedDistribution.distribute method taking as it's dist_func a callable that combines product.administer along with any other intervention-specific logic (e.g. intervention state updates), or a bare product.administer reference.

ckirkman-IDM:
Leaving Product merge/subclassing/etc for a later time.

Issue #499, opened 2026-06-06
MSM docstrings


robynstuart:
Both `AgeMatchedMSM` and `AgeApproxMSM` docstrings say *"Extends
`StructuredSexual` for MSM partnerships"* but they actually extend
`MFNetwork`.

Issue #498, opened 2026-06-06
MSMScaleFreeNetwork has no participation filter


robynstuart:
```python
msm = sti.MSMScaleFreeNetwork(target_mean_degree=5.0)
sim.pars.networks.append(msm)
# Runs, but every post-debut male is now in the MSM pool — not a
# realistic MSM-population share.
```

**Root cause:** `MSMScaleFreeNetwork._get_pool` ([msm.py:170-173](/home/robyn/stisim/stisim/networks/msm.py#L170-L173))
returns `over_debut & male` with no participant filter:

```python
def _get_pool(self):
    """Eligible agents the kernel operates on. Default: post-debut males."""
    return self.over_debut & self.sim.people.male
```

`AgeMatchedMSM` uses `msm_share` and `set_msm` to flip `self.participant`
off for non-MSM males; `MSMScaleFreeNetwork` does neither. Result: a
sim with 50/50 sex ratio and `n_agents=10_000` puts ~5000 males in
the MSM network, regardless of intended share.

**Expected:** consistent API with `AgeMatchedMSM`. A `msm_share`
parameter (or similar) that controls the fraction of males flagged
as MSM participants.

**Fix sketch:**

```python
class MSMScaleFreeNetwork(BaseNetwork):
    def __init__(self, pars=None, name=None, **kwargs):
        super().__init__(name=name)
        self.define_pars(
            msm_share=ss.bernoulli(p=0.015),  # match AgeMatchedMSM default
            target_mean_degree=2.0,
            target_mean_dur=ss.years(2),
            max_edge_dur=ss.years(10),
            phi=1.0,
        )
        self.update_pars(pars, **kwargs)
        ...

    def set_network_states(self, upper_age=None):
        super().set_network_states(upper_age=upper_age)
        # Flip participant off for non-MSM males + all females.
        ppl = self.sim.people
        m_uids = ppl.male.uids
        self.participant[m_uids] = self.pars.msm_share.rvs(m_uids)
        self.participant[(~ppl.male).uids] = False

    def _get_pool(self):
        return self.over_debut & self.sim.people.male & self.participant
```

Verified workaround (currently used as a local subclass in
`/tmp/msm_sustainability_check.py`):

```python
class ScaleFreeMSMShare(sti.MSMScaleFreeNetwork):
    def __init__(self, msm_share=0.005, **kwargs):
        super().__init__(**kwargs)
        self._msm_share = float(msm_share)

    def init_post(self):
[...]

Issue #497, opened 2026-06-06
AgeMatchedMSM crashes at sim init


robynstuart:
```python
msm = sti.AgeMatchedMSM(msm_share=ss.bernoulli(p=0.005))
sim.pars.networks.append(msm)
sim.init()
sim.run()
# ValueError: Invalid entries for age difference preferences.
# Raised from stisim/networks/mf.py:163
```

**Root cause:** `AgeMatchedMSM` extends `MFNetwork`, which initialises
F–M age-difference preference (`age_diff_pars`) tables in
`get_age_risk_pars` ([mf.py:140-165](/home/robyn/stisim/stisim/networks/mf.py#L140-L165)).
The default `age_diff_pars` table is keyed by `(sex_of_p1, risk_group_p1, risk_group_p2)`
or similar F–M risk-group combinations. With a male-only pool the
female-side entries are NaN, the check at line 162–164 fires:

```python
if np.isnan(scale).any() or np.isnan(loc).any():
    errormsg = 'Invalid entries for age difference preferences.'
    raise ValueError(errormsg)
```

**Expected:** MSM matching should not require F–M age difference
preferences. `AgeMatchedMSM.match_pairs` sorts males by age (line 66-72)
and never consults `age_diffs` — but the inherited init still does.

**Fix sketch (one of):**
- Override `set_network_states` in `AgeMatchedMSM` to skip the F–M
  age-diff init for MSM-only networks.
- Or initialise the M-M entries of `age_diff_pars` to safe defaults
  (`loc=0.0, scale=2.0`) inside `AgeMatchedMSM.__init__`.

Issue #496, opened 2026-06-06
AgeApproxMSM doesn't accept `msm_share`


robynstuart:
```python
import starsim as ss
import stisim as sti
msm = sti.AgeApproxMSM(msm_share=ss.bernoulli(p=0.005))
# ValueError: 1 unrecognized arguments for msm: msm_share
```

**Root cause:** `AgeApproxMSM.__init__` forwards `**kwargs` directly
to `MFNetwork.__init__`, which doesn't define `msm_share`. Only
`AgeMatchedMSM.__init__` calls `self.define_pars(msm_share=...)`.

```python
# AgeMatchedMSM (works):
def __init__(self, pars=None, **kwargs):
    super().__init__(name='msm')
    self.define_pars(msm_share=ss.bernoulli(p=0.015))
    self.update_pars(pars=pars, **kwargs)

# AgeApproxMSM (broken):
def __init__(self, **kwargs):
    super().__init__(name='msm', **kwargs)  # no msm_share defined here
```

**Expected:** API consistency between the two MSM variants. If
`msm_share` is the canonical knob on `AgeMatchedMSM` it should also
be available on `AgeApproxMSM`.

**Fix sketch:** Mirror `AgeMatchedMSM`'s init in `AgeApproxMSM`:

```python
def __init__(self, pars=None, **kwargs):
    super().__init__(name='msm')
    self.define_pars(msm_share=ss.bernoulli(p=0.015))
    self.update_pars(pars=pars, **kwargs)
```

And add a `set_msm` method (copying `AgeMatchedMSM.set_msm`) so the
participant filter is applied at init.

PR #502, opened 2026-06-12
Fix MSM network participation filter + init crashes (#496–#499)

robynstuart:
Fixes four MSM-network issues (milestone v1.5.7): #496, #497, #498, #499.

## #496 — `AgeApproxMSM` doesn't accept `msm_share`
- `__init__` now defines `msm_share` and `update_pars`, mirroring `AgeMatchedMSM`; adds `set_network_states` + `set_msm`.
- While here, fixed `match_pairs` so the network actually runs: it had a `(self, ppl)` signature (called by `add_pairs` with no args), called `self.over_debut()` on a property, and used `p1 = m_eligible.uids[ind_p1]` / `group2[ind_p2]` which mismatched the split groups and could `IndexError` on odd counts. It now pairs `group1`↔`group2` by age preference and raises `NoPartnersFound` when fewer than two males are looking.

## #497 — `AgeMatchedMSM` crashes at sim init
- Root cause: `set_network_states` overrode the base to call *only* `set_msm`, so `set_risk_groups` never ran and `risk_group` stayed NaN, tripping the NaN check in `get_age_risk_pars` during `init_post`.
- Fix: `set_network_states` now calls `super().set_network_states()` (debut, risk groups, concurrency) before `set_msm`.

## #498 — `MSMScaleFreeNetwork` has no participation filter
- Added `msm_share` (default `ss.bernoulli(p=0.015)`, matching `AgeMatchedMSM`), a `set_network_states`/`set_msm` pair, and `_get_pool` now filters on `self.participant`.

## #499 — docstrings
- `AgeMatchedMSM` and `AgeApproxMSM` now correctly say they extend `MFNetwork` (not `StructuredSexual`).

## Beyond the literal issues
- **`match_pairs` now filters on `self.participant`** in both MF-based MSM variants — previously `msm_share`/`set_msm` set `participant` but matching ignored it, so the knob was inert.
- **Tests:** `test_msm_network` passed `networks=msm` (a lone instance), which `sti.Sim.process_networks` silently replaced with the default `StructuredSexual` — so it wasn't testing MSM at all. Now uses `networks=[msm]` with a non-trivial `msm_share`. The scalefree kernel/dynamics tests route through a full-participation helper (the new 0.015 default would otherwise leave a degenerate pool at small `n_agents`); added `test_msm_share_filters_pool`.

All `test_scalefree_network.py`, `test_networks.py`, `test_mfnetwork.py`, `test_match_methods.py` pass.

### DIFF
--- a/stisim/networks/msm.py
+++ b/stisim/networks/msm.py
@@ -19,7 +19,7 @@ __all__ = ['AgeMatchedMSM', 'AgeApproxMSM', 'MSMScaleFreeNetwork']
 class AgeMatchedMSM(MFNetwork):
     """Men-who-have-sex-with-men network using exact age-sorted matching.
 
-    Extends :class:`StructuredSexual` for MSM partnerships. Eligible males
+    Extends :class:`MFNetwork` for MSM partnerships. Eligible males
     are sorted by age and paired sequentially so that partners have similar
     ages. The ``msm_share`` parameter controls what fraction of males
     participate.
@@ -40,6 +40,7 @@ class AgeMatchedMSM(MFNetwork):
         return
 
     def set_network_states(self, upper_age=None):
+        super().set_network_states(upper_age=upper_age)  # debut, risk groups, concurrency
         self.set_msm(upper_age=upper_age)
         return
 
@@ -55,7 +56,7 @@ class AgeMatchedMSM(MFNetwork):
         # Find people eligible for a relationship
         active = self.over_debut
         underpartnered = self.partners < self.concurrency
-        m_eligible = active & ppl.male & underpartnered
+        m_eligible = active & ppl.male & underpartnered & self.participant
         m_looking = self.pars.p_pair_form.filter(m_eligible.uids)
 
         if len(m_looking) == 0:
@@ -123,6 +124,7 @@ class MSMScaleFreeNetwork(BaseNetwork):
     def __init__(self, pars=None, name=None, **kwargs):
         super().__init__(name=name)
         self.define_pars(
+            msm_share=ss.bernoulli(p=0.015),  # fraction of males in the MSM pool
             target_mean_degree=2.0,
             target_mean_dur=ss.years(2),
             max_edge_dur=ss.years(10),
@@ -162,6 +164,16 @@ class MSMScaleFreeNetwork(BaseNetwork):
             )
         return
 
+    def set_network_states(self, upper_age=None):
+        super().set_network_states(upper_age=upper_age)  # set debut
+        self.set_msm(upper_age=upper_age)
+        return
+
+    def set_msm(self, upper_age=None):
+        _, m_uids = self._get_uids(upper_age=upper_age)
+        self.participant[m_uids] = self.pars.msm_share.rvs(m_uids)
+        return
+
     def _get_rng(self):
         """Step-local numpy Generator. Not CRN — see class docstring."""
         seed = int(self.sim.pars.rand_seed) + 2003 + int(self.ti)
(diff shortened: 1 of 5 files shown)


stisim/networks/msm.py
```diff
@@ -123,6 +124,7 @@ class MSMScaleFreeNetwork(BaseNetwork):
     def __init__(self, pars=None, name=None, **kwargs):
         super().__init__(name=name)
         self.define_pars(
+            msm_share=ss.bernoulli(p=0.015),  # fraction of males in the MSM pool
```
ckirkman-IDM: Not for today, but we could consider the name of this parameter, msm_share just doesn't strike me as obvious like msm_rate, msm_participation, etc do. Unless "share" is standard field lingo.

stisim/networks/msm.py
```diff
@@ -162,15 +164,25 @@ def init_pre(self, sim):
...
         return
 
+    def set_network_states(self, upper_age=None):
+        super().set_network_states(upper_age=upper_age)  # set debut
+        self.set_msm(upper_age=upper_age)
+        return
+
+    def set_msm(self, upper_age=None):
+        _, m_uids = self._get_uids(upper_age=upper_age)
+        self.participant[m_uids] = self.pars.msm_share.rvs(m_uids)
```
ckirkman-IDM: I kinda feel like this self.participant warrants documentation somewhere other than just setting it here, since it is set via post_init()->set_network_states().

stisim/networks/msm.py
```diff
@@ -461,42 +473,57 @@ def add_pairs(self):
...
+    def __init__(self, pars=None, **kwargs):
+        super().__init__(name='msm')
+        self.define_pars(
+            msm_share=ss.bernoulli(p=0.015),
+        )
+        self.update_pars(pars=pars, **kwargs)
+        return
 
-    def match_pairs(self, ppl):
+    def set_network_states(self, upper_age=None):
```
ckirkman-IDM, line 509: Maybe make an issue to explore ways to make it so we don't have to have copies of set_network_states in all msm classes that are identical, to help reduce the same issues cropping up if more are created?

stisim/networks/msm.py
ckirkman-IDM: Same for set_msm().

stisim/networks/msm.py
```diff
@@ -461,42 +473,57 @@ def add_pairs(self):
...
+        super().set_network_states(upper_age=upper_age)  # debut, risk groups, concurrency
+        self.set_msm(upper_age=upper_age)
+        return
+
+    def set_msm(self, upper_age=None):
+        _, m_uids = self._get_uids(upper_age=upper_age)
+        self.participant[m_uids] = self.pars.msm_share.rvs(m_uids)
+        return
+
+    def match_pairs(self):
```
ckirkman-IDM, line 526: Future issue: do we need a separate match_pairs(), given that has been abstracted for MFNetwork? (is there a way or is it desirable to open this code up to using/sharing matching algorithms)?

tests/test_scalefree_network.py
```diff
@@ -60,10 +67,23 @@ def test_get_pool_filters_to_post_debut_males():
     assert debut_ok.all(), 'pool contains pre-debut agents'
 
 
+@sc.timer()
+def test_msm_share_filters_pool():
+    """``msm_share`` controls the fraction of post-debut males in the pool."""
+    full = _make_sim(net=_make_net(msm_share=ss.bernoulli(p=1.0)), n_agents=2_000)
+    part = _make_sim(net=sti.MSMScaleFreeNetwork(msm_share=ss.bernoulli(p=0.2)), n_agents=2_000)
```
ckirkman-IDM: line 74, use _make_net for consistency?

ckirkman-IDM (approved):

stisim/networks/msm.py
robynstuart: agree - renamed

stisim/networks/msm.py
robynstuart: done

stisim/networks/msm.py
robynstuart: https://github.com/starsimhub/stisim/issues/510

stisim/networks/msm.py
robynstuart: https://github.com/starsimhub/stisim/issues/511

tests/test_scalefree_network.py
robynstuart: done!

PR #503, opened 2026-06-12
SyndromicManagement/SymptomaticTesting: take disease names; fix has_cerv_symp state

robynstuart:
Fixes disease-reference handling in the two STITest-symptomatic interventions, plus a `has_cerv_symp` state bug in `SyndromicManagement`.

## Root cause: deep-copied disease refs

`Sim` deep-copies the diseases and interventions lists in **separate** operations, so a disease object stashed in an intervention at construction is copied independently of the one in `sim.diseases` (deepcopy only preserves shared identity *within* a single call). The intervention's `self.diseases` end up pointing to **orphan copies the sim never initializes or steps**:

- `SyndromicManagement` / `SymptomaticTesting` read `d.symptomatic` / `d.treatable` / `d.susceptible` off the orphans → stale, all-False state.
- `SymptomaticTesting` *also* writes diagnostic results (`new_true_pos`, `new_false_pos`, …) onto the disease module → on orphans those writes are **silently dropped** from the sim output.

## Fix

Both classes now take disease **names** (e.g. `diseases=['ng', 'ct', 'tv']`) and resolve them to the live modules via `sim.diseases[name]` in `init_pre` — consistent with the `STITreatment` / `gonorrhea` / `syphilis` convention. The object was only ever a name-carrier; passing a name makes the (non-)linkage honest.

Also in `SyndromicManagement`: `has_cerv_symp` is now a persistent `BoolState` reset each step, instead of a transient `ss.BoolArr` re-created and re-initialized every step.

## Verification

Smoke-tested both: disease names resolve to the live sim modules, cervical defaults to `['ng','ct']`, runs produce care seekers, and `SymptomaticTesting`'s diagnostic results now land on the live disease (`sim.results.ng.new_true_pos > 0`). STI + intervention suites green.

Docs example and `devtest_amr` updated to the string API.

### DIFF
--- a/stisim/interventions/base_interventions.py
+++ b/stisim/interventions/base_interventions.py
@@ -259,6 +259,9 @@ class SymptomaticTesting(STITest):
     Unlike other test classes, this doesn't return positive/negative outcomes, since
     syndromic management doesn't involve reaching a diagnosis.
     Rather, the testing intervention itself contains a linked treatment intervention.
+
+    ``diseases`` is a list of disease names (e.g. ``['ng', 'ct', 'tv']``),
+    resolved against ``sim.diseases`` at init.
     """
 
     def __init__(self, pars=None, treatments=None, diseases=None, disease_treatment_map=None, negative_treatments=None,
@@ -282,9 +285,9 @@ class SymptomaticTesting(STITest):
         )
         self.update_pars(pars, **kwargs)
 
-        # Store treatments and diseases
+        # Store treatments and diseases (diseases are names, resolved in init_pre)
         self.treatments = sc.tolist(treatments)
-        self.diseases = diseases
+        self.diseases = sc.tolist(diseases)
         if disease_treatment_map is None:
             disease_treatment_map = {t.disease: t for t in self.treatments}
         self.disease_treatment_map = disease_treatment_map
@@ -303,6 +306,11 @@ class SymptomaticTesting(STITest):
 
     def init_pre(self, sim):
         super().init_pre(sim)
+        # diseases are passed as names; resolve to the live sim modules. Sim
+        # deep-copies modules, so constructor-passed objects would be orphans
+        # the sim never steps — state reads and result writes below would hit
+        # dead copies.
+        self.diseases = [sim.diseases[name] for name in self.diseases]
         if self.treat_prob_data is not None:
             self.treat_prob = sc.smoothinterp(sim.yearvec, self.treat_prob_data.year.values, self.treat_prob_data.treat_prob.values, smoothness=0)
         return
--- a/stisim/interventions/base_interventions.py
+++ b/stisim/interventions/base_interventions.py
@@ -513,6 +513,7 @@ class SyndromicManagement(STITest):
         self.define_states(
             ss.FloatArr('ti_referred'),
             ss.FloatArr('ti_dismissed'),
+            ss.BoolArr('has_cerv_symp'),
         )
         self.treat_prob_data = treat_prob_data
         self.treated_by_uid = None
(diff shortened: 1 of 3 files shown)


stisim/interventions/base_interventions.py
```diff
@@ -524,9 +525,13 @@ def init_pre(self, sim):
         self.pars.tx_cerv_m.set(p=self.mvals_cerv)
         self.pars.tx_noncerv_f.set(p=self.fvals_noncerv)
         self.pars.tx_noncerv_m.set(p=self.mvals_noncerv)
-        # Resolve cervical diseases: use explicit list, else find 'ng'/'ct' in diseases
-        if not self._cervical_diseases:
-            self._cervical_diseases = [d for d in self.diseases if d.name in ('ng', 'ct')]
+        # Resolve diseases via sim.diseases so state/result references are linked
+        self.diseases = [sim.diseases[d.name] for d in self.diseases]
```
ckirkman-IDM: Grabbing the diseases off of the sim vs using self.diseases passed in from constructor ...

Reason:
"so state/result references are linked"

Question:
Is this due to starsim doing a deep copy of modules when it creates simulations?
Issue path:
```
diseases1 = [make some objects]  # ref1
interventions = [SyndromicManagement(diseases=diseases1)]  # using ref1
sim = stisim.Sim(diseases=diseases, interventions=interventions)  # given ref1, but sim deep-copies ALL
diseases2 = sim.diseases  # ref2

disease3 = interventions[0].diseases  # ref3? If so, not "linked" to sim or any other intervention/module.
```

However, starsim.Sim() (superclass of ss) will deep copy by default all incoming args/kwargs/pars. From experience, does this end up with the final line above (needing re-linking)? If so, this is a very important aspect for enabling/disabling memory sharing (object sharing) that, for example, will need to be considered for logistics (when sharing Supplies).

If not, could you elaborate on the need to re-link, as you put it (for my understanding and to ensure it is needed).

ckirkman-IDM: If this is a non-issue (and is only for my information) than please just affirm the reason for change (for my understanding, please, no changes needed :)) , and then merge the PR at your leisure. (otherwise if not needed to do the "linking" then it would be best to leave it out.

ckirkman-IDM (changes requested):

stisim/interventions/base_interventions.py
robynstuart: Tried this:

```
import starsim as ss
import stisim as sti

ng = sti.Gonorrhea(init_prev=0.1)

class TestIntv(ss.Intervention):
    def __init__(self, disease):
        super().__init__()
        self.mydis = disease       # stash the constructor-passed reference
    def step(self): pass

test_intv = TestIntv(ng)
sim = sti.Sim(diseases=[ng], networks=sti.StructuredSexual(), interventions=[test_intv], n_agents=300, dur=1, verbose=-1)
sim.init()

ng_in_sim = sim.diseases.ng
print('original ng IS sim.diseases.ng:        ', ng is ng_in_sim)
print('test_inv object IS sim.interventions[0]:  ', probe is sim.interventions[0])
print('test_inv.mydis IS sim.diseases.ng:        ', sim.interventions[0].mydis is ng_in_sim)
# Do they observe the same state after a step? mutate sim's disease, read via the stashed ref
sim.run(verbose=-1)
print('same .infected array object:           ', sim.interventions[0].mydis.infected is ng_in_sim.infected)
print('stashed.infected.sum=%d  sim.diseases.ng.infected.sum=%d' % (int(sim.interventions[0].mydis.infected.sum()), int(ng_in_sim.infected.sum())))
```

Got this:
```
original ng IS sim.diseases.ng:        False   # Sim deep-copied the disease
test_inv object IS sim.interventions[0]:  False   # Sim deep-copied the intervention
test_inv.mydis IS sim.diseases.ng:        False   # the link is BROKEN
stashed.infected.sum=0  sim.diseases.ng.infected.sum=9   # stashed ref is stale/un-stepped
```

So in essence: the intervention's refs are orphan copies that the sim never initializes or steps, so reading `d.symptomatic` off them returns stale/all-False state.

Which leads to the follow-on question: would it be better to contruct the intervention with strings -- `test_intv = TestInv('ng')`? What's the advantage of using the actual objects if they aren't linked? The codebase seems unresolved on this. `STITreatment` and all its subclasses use string-based linkage (`diseases=['ng']`). Only `SymptomaticTesting` and `SyndromicManagement` (the two STITest-symptomatic classes) use object-based linkage. 

My sense is that the object is a misleading name-carrier that looks linked but isn't. So I'm going to change it to a string here!

PR #504, opened 2026-06-12
PartnershipFormationAnalzer

### LINKED ISSUE #479: Add analyzers for partnership rates by age

ckirkman-IDM:
This is a draft PR as I have yet to manually review the test code updates. I have checked the rest.

Feel free to comment/suggest on anything, including whether claude's (moderately updated by me) doc strings are great or trash.

---
Update to networks:

Networks now record the time index when relationships end, leaving a value of None if they don't end (before the end of the simulation). This enables massively more efficient memory usage in PartnershipFormationAnalyzer. No reason for relationship termination is recorded (natural breakup vs death).

Relationships count as ended in the ti they end if a "natural breakup" occurs. If termination is by agent death (post-disease-transmission), the recorded termination time index is set to ti+1, keeping with [formation_ti, ended_ti) half-open interval convention. Consistent with counting the number of disease step integrations the relationship was active for.

---

The core of this analyzer stores data into self.results['relationships']. However, several standard "getter" functions were added to utilize this self.results data to allow one analylzer to be used to record relationship information instead of duplicating the data in self.results across each relationship analyzer (shared memory).

The core behavior of using self.results['relationships'] is the most important part, as it allows any relationship analysis to occur post-run (user defined). 

The getters are the part that is most likely to be adjusted by feedback from users:

get_n_partnerships_formed()
get_n_partnerships_formed_per_agent()
get_unique_partners_active_per_agent()


ckirkman-IDM:
From Claude: (a bit wordy!)

# PR Summary — `PartnershipFormationAnalyzer` and network edge-interval tracking

*(changes since `4abd881`)*

## Overview

Introduces `PartnershipFormationAnalyzer` — a single analyzer that records every partnership as an **interval** (`[formation_ti, ti_expired)`) and answers formation / active-partner queries per network, sex, age bin, and trailing time window. It consolidates previously separate analyzers (the old `NPartnersAnalyzer` and partnership-formation logic) into one method-based reporter, and adds the network-side plumbing that records when each edge forms and when it ends.

## Motivation

There was no consistent, low-overhead way to ask "how many partnerships formed / were active, by whom, when." The analyzer stores **one row per distinct edge with its active interval** instead of re-snapshotting the active edge set every timestep — far cheaper in memory/CPU — and exposes results through getters rather than pre-baked result arrays.

## Changes by area

### Networks (`stisim/networks/`)

**`BaseNetwork`** — recording and clearing of expirations are now entirely network-owned (analyzers are optional, read-only consumers):
- New edge-meta column **`formation_ti`** (timestep the edge formed), set by every `add_pairs`.
- **`record_expired`** flag + **`expired_this_ti`** buffer (a dict mirroring the edge-meta layout).
- **`_append_expired(mask)`** — *accumulates* (concatenates) removed edges into the buffer; gated on `record_expired`.
- **`_on_edge_dissolution(active)`** now calls `_append_expired(~active)` (duration/age expiry at loop **step 7**).
- New **`remove_uids` override** — records death/removal-driven removals (loop **step 15**) into the buffer before delegating to `super().remove_uids()` (which performs the actual removal — documented as required). Closes the gap where dead agents' edges were dropped unrecorded.
- New **`finish_step`** (loop **step 14**) — the network **clears its own buffer** each timestep, after analyzers read it (step 13) and before the next step's removals. With no analyzer present, the buffer is still reset each step so nothing accumulates; multiple analyzers can each read the same buffer.
- New class flag **`records_all_expirations`** (default `True`) — declares whether every removal funnels through the recorded paths.

**`MFNetwork`** — `_on_edge_dissolution` calls `super()` first (so expiry recording fires) before its prior-partner / partner-count bookkeeping; `add_pairs` passes `formation_ti`.

**`SWNetwork`, `MSMScaleFreeNetwork`, `StructuredSexual`** — pass `formation_ti` on append; `add_pairs` TODOs note the no-concurrent-duplicate-edge invariant.

**`MSMScaleFreeNetwork`** — sets `records_all_expirations = False` (its in-`add_pairs` Markov deletes bypass *both* recorded paths), so the analyzer warn-skips it; a TODO documents how to make it compatible (route deletes through `_append_expired`).

### Analyzer (`stisim/analyzers.py`)

- **`PartnershipFormationAnalyzer`** stores, in `self.results['relationships'][nw]`, a list of mutable rows:
  `[p1, p2, formation_ti, ti_expired, age_p1, age_p2, female_p1, female_p2]` (one per distinct edge, keyed by `(p1, p2, formation_ti)`).
- **`_indicies_of_rels_in_table_for_nw`** maps `(p1, p2, formation_ti) → row index`, for O(1) expiry stamping; entries are pruned when an edge's expiry is recorded.
- **`step()`** (loop step 13): appends newly-formed edges (rows with `ti_expired = None`); records `ti_expired` for edges sitting in the network's buffer via the read-only **`_record_expired`** (it does *not* clear the buffer — the network does).
- **`finalize()`**: drains the final-step leftovers (deaths removed at the last step 15, never read during the run) and stamps them `ti_expired = final_ti + 1`. Genuinely still-active edges were never buffered, so they keep `ti_expired = None`.
- **`_rel_element_arrs`** materializes the row list into columnar numpy arrays, mapping `ti_expired = None → final_ti + 1`.
- **Network selection** requires a stisim **`BaseNetwork`**-derived sexual network (the analyzer reads BaseNetwork-only edge meta and the `expired_this_ti` buffer). Named missing networks raise `ValueError`; named non-`BaseNetwork` or `records_all_expirations=False` networks **warn and are skipped** rather than crashing mid-run. Duplicate age bins raise at construction.

### Tests & dev tooling
- New `tests/test_partnership_formation_analyzer.py` (14 tests): injection-based value tests (unique-vs-formed, half-open window boundary, overlapping/out-of-range age bins, same-sex networks, multiple networks), construction/selection/warning tests, and **end-to-end sim tests with HIV mortality and `ss.Deaths`** that verify expiry recording and interval consistency.
- `tests/devtests/devtest_network_diagnostics.py` updated to use the merged analyzer (distribution + population-pyramid diagnostics).

## Key semantics & assumptions

- **`ti_expired` definition**: one past the last timestep the edge was present and transmission-capable at the disease transmission step (loop **step 9**). The active interval is half-open `[formation_ti, ti_expired)`, and
  `duration = ti_expired − formation_ti` = **number of active transmission steps** the relationship existed — consistent regardless of when/why it dissolved.
  - Duration/age expiry at R → `ti_expired = R` (recorded same ti).
  - Partner death/removal at T → `ti_expired = T + 1` (the `alive` flag flips at step 10, *after* step 9, so the edge is still active at T; recorded the next ti).
  - Formed-and-ended-by-death the same ti → `duration = 1` (never an empty interval).
- **Still-active at run end** → `ti_expired = None` (right-censored marker; getters treat `None` as active through `final_ti`). "Active transmission step" means the edge existed with both partners alive at step 9 — *not* that a transmission event fired.
- **Per-agent results cover only agents alive at run end.** The `*_per_agent` getters are indexed by `surviving_agent_uids` (`self._final_uids[...]`, snapshotted at the final `step()`), so agents who died/left before the end have no slot and are dropped from those outputs. The aggregate `get_n_partnerships_formed` is **not** so limited — it includes formations by since-departed agents. (Subject must survive; a surviving subject's partner may since have died and still counts.)
- **Single-edge-per-pair-per-ti assumption** (documented in the class docstring): rows are keyed by `(p1, p2, formation_ti)`. If a network ever formed >1 edge between the **same pair at the same ti**, those collide and **expiry data is corrupted** (all-but-one keep `ti_expired = None`); formation *counts* are unaffected. The bundled networks never do this; re-forming the same pair at a *different* ti is fine.
- **`MSMScaleFreeNetwork` is unsupported** (warn-skipped) until its Markov deletes are routed through the recorded paths.

## What you can pull from the analyzer

All getters return `{network_name: result}`; `female` selects the subject sex; `window_months` is a **trailing window in timesteps** ending at the final timestep (`None` ⇒ whole run; a value longer than the run clamps to the whole run).

- **`get_n_partnerships_formed(female, age_bins=None, window_months=None)`** → `{nw: {age_bin: ndarray}}`. Non-unique formation counts bucketed by age-at-formation. With `window_months=None` the inner array is the **full per-timestep series** (length `n_ti`, indexed by ti); with an int it is a **length-1 array** = total formations over the trailing window. Overlapping age bins each count a formation; out-of-range ages are dropped; a same-sex edge contributes both endpoints.
- **`get_n_partnerships_formed_per_agent(female, window_months=None)`** → `{nw: ndarray}` indexed by `surviving_agent_uids`; non-unique formations per agent over the window (a re-formed pair counts each time).
- **`get_unique_partners_active_per_agent(female, window_months=None)`** → `{nw: ndarray}` indexed by `surviving_agent_uids`; count of **unique partners on edges active during the window** (active iff `ti_expired > win_start`; ongoing edges carry `final_ti + 1` and always count). Sex-general (works for same-sex networks), deduped per `(subject, partner)` so a partner reached via multiple active edges counts once.

Raw per-edge intervals are also available directly in `self.results['relationships'][nw]` for custom analysis (e.g. realized-duration distributions; `ti_expired = None` = right-censored / still active at run end).

## Risk / compatibility
- All network recording is gated on `record_expired` (set only when the analyzer tracks a network) and clearing is a per-step buffer reset, so sims **without** the analyzer are behaviorally unchanged. Upstream starsim is untouched. `MSMScaleFreeNetwork` users get a clear warn-and-skip rather than silent/incorrect results.

stisim/interventions/hiv_test.py
robynstuart, line 1: maybe added by accident? seems to be blank

stisim/networks/base.py
```diff
@@ -59,18 +59,27 @@ class BaseNetwork(ss.SexualNetwork):
...
+    # path (e.g. a custom delete inside ``add_pairs``) MUST set this to False.
+    records_all_expirations = True
+
+    def __init__(self, name=None, record_expired: bool = False, **kwargs):
         super().__init__(name=name)
         self.meta.condoms = ss_float
         self.meta.age_p1 = ss_float
         self.meta.age_p2 = ss_float
         self.meta.edge_type = ss_float
+        self.meta.formation_ti = ss_int  # timestep at which the edge was formed
```
robynstuart: slight preference for ti_formation or ti_start for consistency with other ti_* patterns

stisim/networks/base.py
```diff
@@ -59,18 +59,27 @@ class BaseNetwork(ss.SexualNetwork):
...
         self.define_pars(**BasePars())
         self.define_states(
             ss.BoolArr('participant', default=True),
             ss.FloatArr('debut', default=0),
             reset=True,
         )
+        self.record_expired = record_expired
+        self.expired_this_ti = {}  # only utilized if self.record_expired is True
```
robynstuart: consider `just_expired`? very minor but the presence of the `ti` string might lead one to expect it to contain a timestep if not reading closely

stisim/analyzers.py
```diff
@@ -1,753 +1,1250 @@
...
-                ax.set_ylabel('ART coverage')
-                ax.set_title(f'{sex_labels[sex]} by age')
-                ax.set_ylim(0, 1)
-                ax.legend(frameon=False, fontsize=9)
-
-        pl.tight_layout()
-        return fig
-
+"""
+Common analyzers for STI analyses
```
robynstuart, line 2: skipping over all these diffs, which afaics are NOT diffs!

stisim/analyzers.py
```diff
@@ -1,753 +1,1250 @@
...
+                ax.set_ylabel('ART coverage')
+                ax.set_title(f'{sex_labels[sex]} by age')
+                ax.set_ylim(0, 1)
+                ax.legend(frameon=False, fontsize=9)
+
+        pl.tight_layout()
+        return fig
+
+
+class PartnershipFormationAnalyzer(ss.Analyzer):
```
robynstuart, line 756: Maybe this is in there somewhere, but would be nice to have an error raised if someone tried to use this in conjunction with a network that doesn't record formation time

stisim/analyzers.py
```diff
@@ -1,753 +1,1250 @@
...
+            # recording of actual ti_expired.
+            new_rel_mask = np.asarray(nw.edges.formation_ti, dtype=np.int64) == ti
+            if new_rel_mask.any():
+                p1arr = np.asarray(nw.edges.p1, dtype=np.int64)[new_rel_mask]
+                p2arr = np.asarray(nw.edges.p2, dtype=np.int64)[new_rel_mask]
+                form_ti_arr = np.asarray(nw.edges.formation_ti, dtype=np.int64)[new_rel_mask]
+                age1arr = np.asarray(nw.edges.age_p1, dtype=float)[new_rel_mask]
+                age2arr = np.asarray(nw.edges.age_p2, dtype=float)[new_rel_mask]
+                # Female sex (T/F) of agents in the relationships; not assuming p1/p2 sex for sexual network generality.
+                fem1arr = np.asarray(ppl.female[ss.uids(p1arr)], dtype=bool)
```
robynstuart, line 972: Claude has a habit of adding `np.asarray()` to ss.uids when not needed. I think these lines can all be simplified (again, non-urgent)

stisim/analyzers.py
```diff
@@ -1,753 +1,1250 @@
...
+        ``remove_uids`` and clears the buffer itself in ``finish_step`` (step 14),
+        so several analyzers can each read the same buffer, and recording works
+        with no analyzer present. Edges never placed in the buffer (i.e. still
+        active) are left untouched, keeping ``ti_expired = None``.
+        """
+        # grab the expired relationship info
+        expired_rels = nw.expired_this_ti
+        # if at least one relationship expired, record known end dates.
+        if len(expired_rels.get('p1', ())) > 0:
+            p1_arr = np.asarray(expired_rels['p1'], dtype=np.int64)
```
robynstuart, line 1008: as above

robynstuart (commented):
Looks good, thanks @ckirkman-IDM! Just small comments

stisim/interventions/hiv_test.py
ckirkman-IDM: appears to be.

stisim/networks/base.py
ckirkman-IDM: Settled on expired_this_loop

stisim/analyzers.py
ckirkman-IDM: The underlying changes are in BaseNetwork, so you'd have to use a non BaseNetwork or starsim network from outside stisim. Adding a note on the analyzer.

stisim/analyzers.py
ckirkman-IDM: It really does like to do typing. Maybe we should do a cleanup task someday to find these. I bet there a few extra somewhere.

stisim/networks/base.py
ckirkman-IDM: ti_formed it is now.

PR #501, opened 2026-06-12
PartnerNotification stratification + cycle prevention; MF coital-decay & client multiplier

robynstuart:
Builds on #500 (`feat/syph-detectable-state`). **Base will retarget to `rc1.5.7` once #500 merges** — this PR's diff is the partner-notification + network delta only.

## PartnerNotification

- **Edge-type / partner-sex stratification.** `step()` walks current-channel edges preserving edge type and exposes `current_partner_edges` (`{partner_uid: [edge_type, ...]}`). New `pn_rates(rates)` helper builds a per-UID probability callable for `p_notify_current` / `p_attends_current`, stratified by edge type and (optionally) partner sex:
  - `pn_rates({'stable': 0.20, 'casual': 0.10})`
  - `pn_rates({'stable': {'f': 0.80, 'm': 0.50}, 'casual': {'f': 0.50, 'm': 0.25}})`

  Partners reachable via multiple current-channel edges have per-edge probabilities summed and capped at 1. Scalar bernoullis are unchanged (backward compatible). Edge-type stratification is current-channel only — the prior channel returns 0.
- **Cycle prevention** (`prevent_cycles=True`, default ON). A per-agent `last_notifier` blocks the A→B→A cycle while still allowing chains A→B→C. Pass `prevent_cycles=False` for the legacy no-memory behaviour.
- **Diagnostic results** (opt-in): `new_attended_no_sti` (attendees with none of the listed `diseases` infected — false-alarm clinic-time cost) and `new_index_no_sti` (indices over-treated — treated via `index_treatments` but not actually infected). `index_treatments` requires `diseases`.

## MFNetwork

- `stable_act_decay` (per year, default 0.0): linear decay of per-step acts on stable edges as the relationship ages, with a floor. Defaults preserve existing behaviour.
- `client_marital_act_mult` (default 1.0): scales stable-edge acts when the male partner is an FSW client (wife-displacement). Applies only on combined MF+SW networks (e.g. `StructuredSexual`).
- New edge meta fields `acts_baseline` / `start_ti` and an `update_acts()` step hook that recomputes per-edge acts. SW commercial edges are unaffected.

## Notes

- New RNG draws shift the seeded sequence — downstream calibration runs should re-baseline.
- `pn_rates` and the two new diagnostic results don't yet have automated test coverage (`test_pn` covers the scalar/backward-compatible path).

PR #500, opened 2026-06-12
Syphilis trep/non-trep serology states + configurable prevalence denominator

robynstuart:
Syphilis serology dynamics for survey-comparable prevalence, plus two network concurrency knobs that emerged during the Zimbabwe STI calibration.

## Syphilis

- **Treponemal / non-treponemal states.** New `trep` (treponemal RDT positive) and `nontrep` (RPR/VDRL positive) `BoolState`s with realistic dynamics: ~3-week window period before seroconversion, 80% lifelong trep persistence (20% sero-revert), non-trep titre rise at secondary, decline through late latent, and reversion after early-stage treatment. `SyphTx` schedules post-treatment non-trep reversion and clears window-period trep seroconversion.
- **Results.** `trep_prevalence` and `nontrep_prevalence` (overall, by sex, by age×sex), plus disease-stage prevalences: `sexually_transmissible` (primary | secondary | early latent), `symptomatic` (primary | secondary), and `primary`.
- **State consolidation.** Renamed the old `symptomatic` property → `visibly_symptomatic` (a care-seeking concept: requires a visible chancre/rash), and consolidated the duplicative `active` property into `symptomatic`. `active_prevalence` removed; connectors (`hiv_sti`, `gud_syph`) and `SyphilisPlaceholder` now use `symptomatic`.
- **Configurable denominator.** The prevalence denominator (sexually-active adults in an age band) now reads `age_range`, exposed on `BaseSTI`. Set `age_range=[15, 65]` for a 15–64 household-survey denominator — this replaces the earlier hardcoded `*_prevalence_15_64` result variants.
- `dur_early` default 22–24 months (WHO Europe early-latent boundary); latent `rel_trans` decays from the secondary-stage level.

## Network

- `fsw_mf_conc_mult`: scales FSW non-sex-work (MF) concurrency; values <1 mean active sex workers have fewer non-sex-work partners. Default 1.0 = no effect.
- NB-aware per-risk-group concurrency assignment — previously only `lam` was set, which silently no-oped when `concurrency_dist` was `ss.nbinom`.

## Notes

- New RNG draws shift the seeded sequence; downstream calibration runs should re-baseline.

PR #505, opened 2026-06-13
PartnerNotification: edge-type / partner-sex stratification (pn_rates)

robynstuart:
Adds optional edge-type / partner-sex stratification to `PartnerNotification`, independent of the syph (#500) and MSM (#502) work — touches only `base_interventions.py`.

## What

- `PartnerNotification.step()` walks current-channel edges preserving edge type and exposes `current_partner_edges` (`{partner_uid: [edge_type, ...]}`).
- New `pn_rates(rates)` helper builds a per-UID probability callable for `p_notify_current` / `p_attends_current`, stratified by edge type and (optionally) partner sex:
  - `pn_rates({'stable': 0.20, 'casual': 0.10})`
  - `pn_rates({'stable': {'f': 0.80, 'm': 0.50}, 'casual': {'f': 0.50, 'm': 0.25}})`

  Partners reachable via multiple current-channel edges have per-edge probabilities summed and capped at 1.

## Compatibility

- Scalar `p_notify_current` / `p_attends_current` bernoullis are unchanged — `test_pn` passes as-is.
- Edge-type stratification is current-channel only; the prior channel returns 0 (documented in `step()` / `pn_rates`).

## Notes

- This is the cycle-prevention-free PN contribution (the cycle-prevention variant in the closed #501 was not adopted).
- `pn_rates` has no automated test coverage yet (`test_pn` covers the scalar path); smoke-tested manually (both shapes notify as expected).

### DIFF
--- a/stisim/interventions/base_interventions.py
+++ b/stisim/interventions/base_interventions.py
@@ -1033,8 +1033,9 @@ class PartnerNotification(ss.Intervention):
     Each contact is offered notification with probability ``p_notify_<scope>``;
     notified contacts attend follow-up with probability ``p_attends_<scope>``.
     Attendees are scheduled for the supplied ``test`` intervention on the next
-    timestep. Index cases are not notified about themselves; partners reachable
-    via both channels are not double-notified.
+    timestep. Index cases partnered with each other are not notified as
+    partners — they are already in the cascade, so re-notifying is redundant.
+    Partners reachable via both channels are not double-notified.
 
     Args:
         eligibility: Callable ``f(sim) -> uids`` returning index cases (e.g.
@@ -1138,8 +1139,11 @@ class PartnerNotification(ss.Intervention):
 
     def find_partners(self, nw, index_uids):
         """
-        UIDs of partners of `index_uids` on network `nw`, deduplicated and
-        excluding any index case from being notified about themselves.
+        Unique UIDs of partners of ``index_uids`` on network ``nw``. Partner
+        UIDs that are themselves index cases are dropped: index cases
+        partnered with each other are not notified as partners (they are
+        already in the cascade). Used for the prior-partner channel, which
+        does not stratify by edge type.
         """
         in_p1 = np.isin(nw.p1, index_uids)
         in_p2 = np.isin(nw.p2, index_uids)
@@ -1151,29 +1155,34 @@ class PartnerNotification(ss.Intervention):
         return ss.uids(partners)
 
     def _build_partner_edges(self, nw, index_uids):
-        """Return dict {partner_uid: [edge_type_int, ...]} for current-channel
-        edges where one endpoint is an index case. Edges that connect two
-        index cases are excluded — index cases are not notified about
-        themselves.
+        """Return ``{partner_uid: [edge_type_int, ...]}`` for current-channel
+        edges with one endpoint in ``index_uids``. Partner UIDs that are
+        themselves index cases are dropped: index cases partnered with each
+        other are not notified as partners (they are already in the cascade).
+        A partner reachable via several index cases gets one list entry per
+        edge, so the :func:`stisim.pn_rates` helper can sum per-edge rates.
 
-        Edge-type info is exposed via ``self.current_partner_edges`` so that
-        callables passed to ``p_notify_current.p`` / ``p_attends_current.p``
-        can return per-UID probabilities stratified by edge type (see the
-        :func:`stisim.pn_rates` helper).
+        Every stisim network carries an ``edge_type`` column (set on the
+        network base ``meta``), so this is safe for any current network; the
+        edge-type ints map to names via ``nw.edge_types``.
+
+        The result is exposed via ``self.current_partner_edges`` so callables
+        passed to ``p_notify_current.p`` / ``p_attends_current.p`` can return
+        per-UID probabilities stratified by edge type.
         """
         in_p1 = np.isin(nw.p1, index_uids)
         in_p2 = np.isin(nw.p2, index_uids)
         partner_uids = np.concatenate([nw.p2[in_p1], nw.p1[in_p2]])
         edge_types = np.concatenate([nw.edges.edge_type[in_p1],
                                      nw.edges.edge_type[in_p2]])
-        # Drop index-from-index edges
+        # Drop partners that are themselves index cases (index-index edges).
         keep = ~np.isin(partner_uids, index_uids)
         partner_uids = partner_uids[keep]
         edge_types = edge_types[keep]
 
         partner_edges = defaultdict(list)
-        for uid, et in zip(partner_uids, edge_types):
-            partner_edges[int(uid)].append(int(et))
+        for uid, edge_type in zip(partner_uids, edge_types):
+            partner_edges[int(uid)].append(int(edge_type))
         return partner_edges
 
     def step(self):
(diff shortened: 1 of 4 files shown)


stisim/interventions/base_interventions.py
```diff
@@ -1135,33 +1122,53 @@ def notify_attendees(self, uids):
         return
 
     def find_partners(self, nw, index_uids):
+        """Walk edges of `nw` and return ``(partner_uids, edge_types)`` arrays
+        for edges where exactly one endpoint is in ``index_uids``. Edges
+        connecting two index cases drop out — index cases aren't notified
+        about themselves. One entry per edge: partners reachable via multiple
```
ckirkman-IDM: "index cases aren't notified about themselves"

I think what you mean is:

"index cases aren't notified about **each other**"

Which raises the question for me: why not?

stisim/interventions/base_interventions.py
```diff
@@ -1135,33 +1122,53 @@ def notify_attendees(self, uids):
...
+        edges appear multiple times in ``partner_uids``.
+
+        ``edge_types`` is filled with ``-1`` if ``nw.edges`` has no
+        ``edge_type`` column (e.g. :class:`PriorPartners`). ``pn_rates``
+        callables map unrecognized edge_type ints to probability 0, so this
+        gracefully disables edge-type stratification on such networks.
```
ckirkman-IDM: ``pn_rates`` callables

... don't you mean:

"pn_rates method (below)"? (singular)

ckirkman-IDM: I would also be more explicit about this. Setting something to -1 here, and assuming that somewhere else it is converted to 0 can be brittle and harder to read. I am slightly concerned about code clarity, but more about the potential for accidental breakage by not tying this logic together further.

The lightest suggestion I have is to change setting -1 explicitly in find_partners() to a known "no_edge_type" edge type, for example, if there is an Enum of edge types, a value of it could be:

```
class EdgeTypes(Enum)
    NO_EDGE_TYPE: 12  # or some other chosen number
````
... with a comment calling out this setting and noting the conversion to 0 in pn_rates.


Then using this explicit value/distinction ( in the code below in pn_rates(), something like):

```
# comment here linking reason for skipping 
if edge_type == EdgeTypes.NO_EDGE_TYPE:
    continue  # or other logic for setting to 0
```

This way the inline comments and reserved values are tightly bound to each other in both places for clarity (for when we "update" it and hopefully don't break it! :) )

File: stisim/interventions/base_interventions.py [hunk 3]
```diff
@@ -2,6 +2,8 @@
 Define interventions for STIsim
 """
 
+from collections import defaultdict
```
ckirkman-IDM, line 5: Please add unittests (or short tim tests, but claude should be able to hack a harness in to do a unittest, I think) to verify proper behavior for:
- a single network
- two networks
- ... then throwing in a network that should have pn_rates 0

stisim/interventions/base_interventions.py
```diff
@@ -1135,33 +1122,53 @@ def notify_attendees(self, uids):
...
-        if len(partners) == 0:
-            return ss.uids()
-        partners = np.unique(partners)
-        partners = partners[~np.isin(partners, index_uids)]
-        return ss.uids(partners)
+        p1_is_index = np.isin(nw.p1, index_uids)
+        p2_is_index = np.isin(nw.p2, index_uids)
+        # XOR: keep edges where exactly one endpoint is an index. Both-index
+        # edges (True ^ True) and no-index edges (False ^ False) drop out.
+        has_one_index = p1_is_index ^ p2_is_index
```
ckirkman-IDM: I love python ^ for XOR :) (but it only works for True/False which is both slightly sad and ... probably good)

stisim/interventions/base_interventions.py
```diff
@@ -1135,33 +1122,53 @@ def notify_attendees(self, uids):
...
         index_uids = self.eligibility(self.sim)
         if len(index_uids) == 0:
             return
 
-        # Current-partner channel
-        cur_partners = self.find_partners(self._cur_nw, index_uids)
+        # Current-partner channel: walk edges, group edge_types by partner uid.
+        # Callables on p_notify_current.p / p_attends_current.p can read
+        # `self.current_partner_edges` to stratify by edge type.
+        cur_uids, cur_edge_types = self.find_partners(self._cur_nw, index_uids)
```
ckirkman-IDM: cur_uids here is ... slightly harder to read / misleading? to me, since we convert to cur_partners by unique check a moment later. Perhaps rename cur_uids to cur_edge_uids?

stisim/interventions/base_interventions.py
```diff
@@ -1135,33 +1122,53 @@ def notify_attendees(self, uids):
...
-        # Prior-partner channel (skip partners already notified as current)
+        # Prior-partner channel (skip partners already notified as current).
+        # Edge-type stratification is not currently supported for the prior
+        # channel; pn_rates helpers passed here would see an empty
+        # current_partner_edges and return zeros.
         if self._use_previous:
-            prev_partners = self.find_partners(self._prev_nw, index_uids)
-            prev_partners = ss.uids(prev_partners[~np.isin(prev_partners, cur_partners)])
+            prev_uids, _ = self.find_partners(self._prev_nw, index_uids)
+            prev_partners = ss.uids(np.unique(prev_uids)) - cur_partners
```
ckirkman-IDM: ... This prevents agents from being notified two steps in a row?

stisim/interventions/base_interventions.py
```diff
@@ -1182,6 +1189,88 @@ def step(self):
...
+    sex_dependent = isinstance(sample, dict)
+    if sex_dependent:
+        for k, v in rates.items():
+            if not isinstance(v, dict):
+                raise ValueError(
+                    f"pn_rates: inconsistent rates spec; key '{k}' has "
+                    f"non-dict value while another key is a dict."
+                )
+
+    def func(module, sim, uids):
```
ckirkman-IDM, line 1280: Slightly more descriptive variables in this method and light inline comments would aid in clarity here.

ckirkman-IDM (changes requested):
Already discussed in our standup today, so hopefully no surprises :)

robynstuart:
Addressed in fb00363 (also merged rc1.5.8 into the branch). Going through the comments:

- **"index cases aren't notified about each other" / why not?** Fixed the wording. By design: an index case is already in the cascade (just diagnosed / eligible), so notifying it *as a partner* is redundant — both endpoints of an index–index edge are dropped. Rationale now stated in the class and `find_partners`/`_build_partner_edges` docstrings.
- **`-1` sentinel / EdgeTypes enum.** The refactor removes the sentinel entirely. `find_partners` no longer touches edge types (prior channel), and `_build_partner_edges` reads `nw.edges.edge_type` directly — every stisim network carries an `edge_type` column (set on the network base `meta`), so there's no missing-column case to guard. Noted in the docstring.
- **`pn_rates` callables (singular).** Comments now refer to *the `pn_rates` helper*.
- **`^` XOR / `cur_uids` rename.** Both gone in the refactor — `step` no longer builds `cur_uids`, and the XOR edge-selection was replaced by the `_build_partner_edges` split.
- **"prevents notification two steps in a row?"** No — the `prev_partners - cur_partners` dedup only prevents double-notification *within a single step* across the two channels. Re-notification across timesteps is governed by the caller's `eligibility` function; `ti_notified` is recorded but not used internally to suppress. Clarified in the inline comment.
- **Descriptive vars + comments in `pn_rates`.** Renamed (`et`→`edge_type`, `spec`→`rate`, `out`→`probs`, `int_to_name`→`edge_type_to_name`) and added inline comments for the empty-channel, unmatched-type, and capping paths.
- **Unit tests.** Added `test_pn_rates` (the probability callable: single/two edge types, multi-edge summing + cap, unmatched-type → 0, empty channel → 0, per-sex rates, inconsistent-spec guard) and `test_pn_rates_sim` (through `step()`: a single current network, a network whose edge types don't match → 0 notifications, and a current+prior two-network sim). Kept them in `test_hiv_interventions.py` alongside the existing scalar `test_pn` rather than a new file.

On test organization more broadly (file/name proliferation): opened #522 to track a proper topic-based restructure rather than do it ad hoc here.

stisim/interventions/base_interventions.py
robynstuart: added

stisim/interventions/base_interventions.py
robynstuart: done

tests/baseline.yaml
```diff
@@ -188,7 +188,7 @@ low_cd4_testing_new_diagnoses: 124.24480082987552
 low_cd4_testing_new_tests: 124.24480082987552
 maternalnet_n_edges: 359398.7938672199
 n_alive: 12736727.974939832
-n_female: 0.0
+n_female: 6580480.923686722
```
robynstuart, line 191: this was wrong before :)

PR #506, opened 2026-06-15
MFNetwork: optional coital-decay + client-husband marital multiplier

robynstuart:
Two new knobs on the heterosexual network:

- MFPars.stable_act_decay (per year, default 0.0): linear decay of per-step acts on stable edges as the relationship ages. Defaults preserve existing behaviour. Empirically, monthly coital frequency in long-running couples is ~50% of newlywed baseline by year 5-7; values around 0.07-0.10/yr with a floor near 0.3-0.4 are reasonable.
- SWPars.client_marital_act_mult (default 1.0): scalar applied to stable-edge per-step acts when the male partner (p1) is a client of an FSW. Captures wife-displacement — extramarital activity often substitutes for marital activity rather than adding to it. Acts only on networks that combine MF + SW (e.g. StructuredSexual).

Two new edge meta fields:
- acts_baseline (preserves the original per-edge act count assigned at edge formation)
- start_ti (timestep when edge was formed; used for decay-age math)

New BaseNetwork.update_acts() step hook recomputes per-edge acts from acts_baseline * _compute_act_multiplier(). The base override returns None (no-op); MFNetwork applies the duration-decay; StructuredSexual composes that with the client-husband multiplier. SWNetwork's commercial edges (edge_type='sw') are unaffected.

Smoke test: 10-yr sim with decay=0.10, floor=0.3, client_mult=0.5 gives stable mean_acts 2.67 vs baseline 5.87 (55% reduction); casual edges unchanged at 6.01; SW edges unchanged at 4.61.

### DIFF
--- a/stisim/networks/base.py
+++ b/stisim/networks/base.py
@@ -65,6 +65,12 @@ class BaseNetwork(ss.SexualNetwork):
         self.meta.age_p1 = ss_float
         self.meta.age_p2 = ss_float
         self.meta.edge_type = ss_float
+        # acts_baseline preserves the per-edge act count assigned at edge
+        # formation; per-step `acts` may be a decayed multiple of it
+        # (see ``update_acts``). start_ti is the timestep when the edge
+        # was created — used to compute edge age for the decay.
+        self.meta.acts_baseline = ss_float
+        self.meta.start_ti = ss_int
         self.define_pars(**BasePars())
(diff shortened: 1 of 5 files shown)


robynstuart:
Not to be merged yet - will refactor following https://github.com/starsimhub/stisim/pull/504 review / merge

PR #527, opened 2026-07-05
Update HIV tests

cliffckerr:
Updating Starsim to v3.5.0 uncovered some tests that failed stochastically when the random number stream changed. This fixes that and has some other minor tidying.


ckirkman-IDM (approved):
lgtm

PR #536, opened 2026-07-10
Fixing infection logging

dtcitron:
I am working on allowing ART to vary transmission probability depending on whether an individual is adhering to ART or not. I wanted to count the number of people infected by each subgroup of ART users, but Claude realized this wasn't possible to do yet in HIVSim.

Here's Claude's explanation of the fix:

"HIV.set_prognoses never called super(), breaking ss.infection_log()"

"HIV overrides Disease.set_prognoses() entirely for its natural-history setup and never called super(), so the infection log's append hook (the only place source/target transmission pairs get recorded) never fired for HIV specifically. Other diseases (e.g. SEIS-based Gonorrhea) already call super() and were unaffected.

Confirmed via test: infection log entries for a 1000-agent HIV sim went from 0 to 259 after the fix, enabling downstream analysis like attributing onward transmission to a source subgroup (e.g. on-ART vs off-ART)"


robynstuart:
@dtcitron the tests are failing on this branch and it's not obvious to me why... could you please take a look?

PR #556, opened 2026-07-18
New model for ART adherence

dtcitron:
- [x] Add EffectiveART/NonSuppressiveART tracking, to enable 90-90-90 metrics tracking
- [x] Allow transmission to vary by EffectiveART/NonSuppressiveART status
- [x] Allow mortality to vary by EffectiveART/NonSuppressiveART status, as well as age, sex, and duration on ART, following ARTMortalityTable from EMOD-HIV
- [x] Add new diagram to the docs to illustrate how the ART adherence stuff can be used to change mortality

Changes to ART adherence. EffectiveART vs NonSuppressiveART allows us to differentiate whether or not an individual adheres to ART. ART status is now one of four things: `art_naive` (replacing never_art) `on_effective_art`, `on_nonsuppressive_art`, and `art_discontinued`. When initiating we pass a new argument `p_effective_art` (defaults to 1.0) which specifies what fraction of individuals who initiate ART reach `on_effective_art status`

ART adherence affects transmission rates - When initiating, the transmission blocking efficacy is now respectively either `effective_art_efficacy=0.96` or `nonsupp_art_efficacy=0.35`. 

ART adherence affects mortality rates - we have borrowed a lookup table from EMOD-HIV, so now mortality varies by Age, Sex, ART adherence, and time since initiating ART.

### DIFF
--- a/docs/user_guide/diseases/hiv.md
+++ b/docs/user_guide/diseases/hiv.md
@@ -14,13 +14,13 @@ HIV in STIsim is modeled with CD4-based disease progression through acute, laten
                                ▼
                         ┌─────────────┐
                         │    Acute    │  CD4 declines from ~800 to ~500
-                        │  (~3 mo)   │  Transmissibility: 6x baseline
+                        │  (~3 mo)    │  Transmissibility: 6x baseline
                         └──────┬──────┘
                                │
                                ▼
                         ┌─────────────┐
                         │   Latent    │  CD4 stable at ~500
-                        │  (~10 yr)  │  Transmissibility: 1x baseline
+                        │  (~10 yr)   │  Transmissibility: 1x baseline
                         └──────┬──────┘
                                │
                     ┌──────────┴──────────┐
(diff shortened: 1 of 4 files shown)


dtcitron:
I had a few questions about these new features and the way they have been integrated with existing code:

1. The code passes the unit tests relevant to transmission and mortality, are there additional checks we should run? I have created a [notebook](https://github.com/dtcitron/HIVSim_notes/blob/main/Testing_new_ART_transmission_and_mortality.ipynb) which runs demos of the new model with spot-checks showing that the code seems to do the things I engineered it to do - until we apply this to a model I won't have a clear sense of exactly how much it will affect how the model behaves.
2. Currently I have opted to hard-code ARTMortalityTable (which does not have a citation in the docs yet) into hiv.py - is this good practice? I will open a p0 issue for this.
3. I realize that the dynamical model governing how CD4 count changes in HIVSim is probably different than the one governing the respective process in EMOD. Should I double-check how different these are, in case that really affects the timing of things.
4. Mortality while on ART was basically zero before. It is much higher now - are there other parts of the code which need to change to accommodate this?

docs/user_guide/diseases/hiv.md
```diff
@@ -155,10 +187,18 @@ HIV in STIsim is modeled with CD4-based disease progression through acute, laten
...
 | Parameter | Default | Description |
 |-----------|---------|-------------|
-| `art_efficacy` | 0.96 | ART efficacy at reducing transmission |
+| `effective_art_efficacy` | 0.96 | Transmission efficacy of effective (virally-suppressive) ART |
+| `nonsupp_art_efficacy` | 0.35 | Transmission efficacy of non-suppressive ART |
 | `time_to_art_efficacy` | 6 months | Time to reach full ART efficacy (linear ramp) |
+| `p_effective_art` | bernoulli(1.0) | Probability a newly-initiated agent achieves viral suppression |
 | `art_cd4_growth` | 0.1 | Logistic growth rate for CD4 reconstitution on ART |
 | `dur_on_art` | lognorm(3 yr, 1.5 yr) | Duration on ART before dropout |
+| `rel_art_mortality_effective` | 0.25 | Fraction of the off-ART CD4-based rate retained on effective ART, both sexes (see "On-ART mortality" above) |
```
robynstuart, line 196: we can work on this later, but i had to read this a few times to make sense of it. I understand it to mean that people on suppressive ART, by default, have their HIV death rates scaled by 1/4. Maybe note here that this is for males and that females have an additional scale factor applied

docs/user_guide/diseases/hiv.md
```diff
@@ -155,10 +187,18 @@ HIV in STIsim is modeled with CD4-based disease progression through acute, laten
...
+| `rel_art_mortality_unsupp_m` | 0.7 | Fraction of the off-ART CD4-based rate retained on non-suppressive ART, males |
+| `rel_art_mortality_unsupp_f` | 0.35 | ...females — chosen so the non-suppressive/effective mortality ratio is ~2x higher for men (2.8x) than women (1.4x) |
+| `rel_death_f` | 0.74 | Additional multiplier for females, on ART (applies equally to effective and non-suppressive; doesn't affect the ratio above) |
+| `art_death_age` | 4 age bins | `(age_lo, age_hi, mult)` list |
```
robynstuart, line 200: Is this something users would be expected to touch? If so will need more instructions here.

docs/user_guide/diseases/hiv.md
```diff
@@ -155,10 +187,18 @@ HIV in STIsim is modeled with CD4-based disease progression through acute, laten
...
+| `art_death_dur` | None | Optional `(dur_lo, dur_hi, mult)` list, days since `ti_art`; off by default |
```
robynstuart, line 201: As above

stisim/diseases/hiv.py
```diff
@@ -59,15 +62,45 @@ def __init__(self, **kwargs):
 
         # Treatment effects
         self.art_cd4_growth = 0.1  # Unitless parameter defining how CD4 reconstitutes after starting ART - used in a logistic growth function
-        self.art_efficacy = 0.96  # Efficacy of ART
+        self.effective_art_efficacy = 0.96  # Transmission efficacy of effective (virally-suppressive) ART
+        self.nonsupp_art_efficacy = 0.35  # Transmission efficacy of non-suppressive ART
         self.time_to_art_efficacy = ss.months(6)  # Time to reach full ART efficacy (in months) - linear increase in efficacy
+        self.p_effective_art = ss.bernoulli(p=1.0)  # Probability that a newly-initiated agent achieves viral suppression (vs. non-suppressive ART)
```
robynstuart, line 68: should we use a different value here since not everyone will achieve suppression?

stisim/diseases/hiv.py
```diff
@@ -59,15 +62,45 @@ def __init__(self, **kwargs):
...
+        # higher for men than for women (0.7/0.25=2.8 vs. 0.35/0.25=1.4) -- rel_death_f below is a separate, adherence-independent
+        # sex multiplier that cancels out of this ratio, so it doesn't affect the 2x relationship.
+        self.rel_death_f = 0.74  # Additional multiplier for females, on ART (applies equally to effective and non-suppressive)
+        self.art_death_age = [  # (age_lo, age_hi, mult), like rel_sus_age
+            (0, 25, 1.0),
+            (25, 35, 1.10),
+            (35, 45, 1.21),
+            (45, 125, 1.32),
+        ]
+        self.art_death_dur = None  # (dur_lo, dur_hi, mult) list, days since ti_art; None = no-op.
```
robynstuart, line 94: Should we provide an example of how you would use this if you wanted to?

stisim/diseases/hiv.py
robynstuart, line 677: minor comment, but usually this goes at the beginning and then you put in class-specific custom code to override whatever the parent class does. Any particular reason for the placement at the end here?

tests/baseline.yaml
```diff
@@ -1,131 +1,134 @@
...
-hiv_n_infected_f_0_15: 42367.47708298754
-hiv_n_infected_f_15_20: 8034.49712033195
-hiv_n_infected_f_20_25: 62867.86921991701
-hiv_n_infected_f_25_30: 87758.24431950209
-hiv_n_infected_f_30_35: 66864.410313278
-hiv_n_infected_f_35_50: 79910.11440041492
-hiv_n_infected_f_50_65: 3375.3170892116177
+hiv_n_acute: 7744.592585062241
+hiv_n_art_discontinued: 9111.285394190872
+hiv_n_art_naive: 12423713.90671577
```
robynstuart, line 42: this seems very large. Do we want `hiv_n_art_naive` to only count HIV+ people who are ART naive? It seems like it might be counting everyone at the moment. What's the standard here?

tests/benchmark.yaml
```diff
@@ -8,5 +8,5 @@ parameters:
     value: 20.0
   n_agents: 2000
 time:
-  initialize: 0.4
-  run: 2.203
+  initialize: 0.328
+  run: 2.13
```
robynstuart: faster!

robynstuart (commented):
Looks good! I will rebase onto v1.5.11 so we can release this week. Please go ahead and merge once you've reviewed the comments & resovled any conflicts

stisim/diseases/hiv.py
dtcitron: This should be configurable, I agree. I'm not sure what happened, I will investigate this.

dtcitron: Should look into finding a better default value.

stisim/diseases/hiv.py
dtcitron: This was left in as a possible feature to be added later, in case we want history-dependent mortality rates that depend on the time since initiating ART in addition to age/sex/cd4 count.

For now this should be left at the None default

docs/user_guide/diseases/hiv.md
dtcitron: Creating an issue to clarify this section of documentation

PR #559, opened 2026-07-23
feat(hiv): expose CD4-mortality rate array and add rel_death scalars

robynstuart:
The CD4-stratified HIV mortality table and its off-ART-only filter were hardcoded inside `HIV.make_p_hiv_death` and `HIV.step`. Coming out of a Zimbabwe calibration exercise where the sim under-counted AIDS deaths by 3-4x after ART scale-up, this makes both aspects tunable.

Changes to `HIVPars`:
- `cd4_death_bins`: descending CD4 bin edges (default matches old hardcoded value)
- `cd4_death_rates`: per-year mortality for each bin (default matches old hardcoded value)
- `rel_death`: scale on all HIV deaths (default 1.0)
- `rel_death_on_art`: additional scale for on-ART agents (default 0.5, from Marston 2007 / Kranzer 2013 African ART cohorts showing on-ART mortality ~30-50% of untreated at each CD4 level)

Behaviour change: on-ART agents are now included in the stochastic HIV-death channel at 0.5x their untreated CD4-bin rate, where before they were excluded entirely. Set `rel_death_on_art=0` to preserve the pre-change behaviour (on-ART agents die only via CD4=0).

Added `test_rel_death_scales_hiv_mortality` to tests/test_hiv.py.


robynstuart:
superseded by https://github.com/starsimhub/stisim/pull/556

PR #561, opened 2026-07-24
Simplify ART mortality to an age/CD4 function, integrate rel_death

robynstuart:
## Summary

Follow-up to #556. Replaces the `(dur x age x cd4) x4` EMOD ARTMortalityTable lookup with a small formula:

```
rate = rel_death * art_death_rate * rel_death_unsupp? * rel_death_f? * age_mult(age) * cd4_mult(cd4)
```

- Drops the duration-since-ART-initiation axis: STIsim's own CD4-reconstitution curve already makes CD4 fall with time-on-ART, so it reproduces the *direction* of the duration effect, but only ~15-30% of its *magnitude* (checked empirically against the original EMOD table). This is a deliberate simplification, documented in `docs/user_guide/diseases/hiv.md`.
- `art_death_dur` ships as an opt-in no-op (`None` by default) with the EMOD-consistent duration bins/multipliers given in a comment, for anyone who wants closer fidelity to the original table.
- All values (`art_death_rate`, `rel_death_unsupp`, `rel_death_f`, `art_death_age`, `art_death_cd4`) were fit from the same EMOD-HIV ARTMortalityTable data via log-additive decomposition, cross-validated against the real table cells.
- On-ART mortality is no longer opt-in (`use_art_mortality_table` removed) — agents on ART are no longer immune to HIV death by default.

Also integrates `feat/hiv-death-rate-pars` (#559)'s `cd4_death_bins`/`cd4_death_rates` (tunable off-ART CD4 mortality table) and `rel_death` (now a *global* scalar applied to both off-ART and on-ART hazards). Drops that branch's `rel_death_on_art`, since it's superseded by the dedicated on-ART function here.

Naming sweep for consistency with `rel_death`/`rel_sus`: `art_mort_supp` -> `art_death_rate`, `rel_art_mort_unsupp` -> `rel_death_unsupp`, `rel_art_mort_f` -> `rel_death_f`, `art_mort_age`/`art_mort_cd4` -> `art_death_age`/`art_death_cd4`.

Also fixes two pre-existing flaky tests (`test_cd4_counts_decline_untreated`, `test_aids_transmission_is_higher_than_latent`) that used `n_agents` too small to reliably produce the outcome they were checking for under the current starsim RNG.

## Test plan
- [x] `pytest tests/test_hiv.py` — 23 passed
- [x] New `test_rel_death_scales_hiv_mortality`, checking `rel_death` scales both off-ART and on-ART death channels monotonically
- [x] Manually verified `get_art_mortality_hazard` runs cleanly with `art_death_dur` both unset (default) and set (EMOD-restoring values)
- [x] Cross-validated the fitted `art_death_rate`/`age_mult`/`cd4_mult`/`rel_death_unsupp`/`rel_death_f` against real cells from the original EMOD table (ratios within ~0.84-1.20x across 24 spot checks)


dtcitron:
As far as I can tell, this fits all of the correct qualitative behaviors that we want to see - looks as though mortality while on ART is higher for males than females; increases with age; increases while CD4 count goes down.

What remains is to figure out how to make this quantitatively comparable to the EMOD-HIV lookup tables. Looking closer at those tables, there are a few idiosyncrasies so I don't think we need to copy-paste them exactly. The extra axis of those tables - duration on ART - we no longer use so it's hard to make a direct quantitative comparison between the new lookup function against the old lookup tables. I have attached two images, one showing a comparison between the tables at <6 months and at 1-2 years - the comparisons appear different and agree/disagree in different places depending on the duration.

Two questions:
- In the docs it says the baseline art_death_rate = 0.0186, but in the main code the baseline art_death_rate is 0.00554 - which of these values should we use? 
- The ratio between mortality among unsuppressed/suppressed for males is 11.4x and ~2.3x for females in EMOD-HIV, but the ratio is 2.03 for both males and females in the new lookup tables. Once we determine if we need to make other quantitative changes, I can change the default ratios to be different by sex.

@robynstuart can you say more about how you reduced the ARTMortality tables down to this new function? Are there any other quantitative adjustments we should make here before committing?

comparison_at_1-2Yr
comparison_at_6Mo

dtcitron:
I have now updated this branch to adjust a few things:

1. We now have art mortality bounded from above by mortality while off art - nobody on ART should have higher mortality risk than someone on ART
2. Mortality while on nonsuppressive ART relative to effective ART is now higher for men than for women
3. (Updated docs to reflect this)

Remaining to do: need to review literature and make sure that numerically the mortality is still empirically supported, and need to add references for parameter values to the docs.

image

